PR #3, opened 2023-08-04
Add input validation and UDP testing

SalDaniele:
This commit introduces some validation of user input and sets reasonable defaults on option arguments.

It also adds the option for iperf UDP testing.

### DIFF
--- a/common.py
+++ b/common.py
@@ -2,6 +2,12 @@ import jinja2
 from enum import Enum
 
 
+
+class TestType(Enum):
+    IPERF_TCP  = 1
+    IPERF_UDP  = 2
+    HTTP       = 3
+
 class PodType(Enum):
     NORMAL     = 1
     SRIOV      = 2
--- a/common.py
+++ b/common.py
@@ -2,6 +2,8 @@ import jinja2
 from enum import Enum
 
 
+FT_BASE_IMG = "quay.io/wizhao/ft-base-image:0.9"
+TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:0.3"
 
 class TestType(Enum):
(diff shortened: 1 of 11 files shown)


iperf.py
```diff
@@ -122,14 +126,36 @@ def stop(self):
...
         if r.returncode != 0:
             logger.info(r)
-
         data = json.loads(r.out)
+        if self.test_type == TestType.IPERF_TCP:
+            self.print_tcp_results(data)
+        if self.test_type == TestType.IPERF_UDP:
+            self.print_udp_results(data)
+    
+    def print_tcp_results(self, data: dict):
```
wizhaoredhat, line 135: I guess the TCP results needs work as well to display as much info as UDP results.

testSettings.py
```diff
@@ -59,7 +60,7 @@ def _is_same_node_test(self, test_id: int) -> bool:
 
     def _server_test_to_persistent(self, test_id: int) -> bool:
         #TODO: add logic to determine when this is required
-        return False
+        return True
```
wizhaoredhat, line 63: Let's try to run the tests without persistent at first. I think it is the harder option.

common.py
```diff
@@ -2,6 +2,14 @@
 from enum import Enum
 
 
+FT_BASE_IMG = "quay.io/wizhao/ft-base-image:0.9"
+TFT_BASE_IMG = "quay.io/wizhao/tft-tools:0.3"
```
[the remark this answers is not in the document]
SalDaniele: Updated

iperf.py
SalDaniele: Yes, I can add this in a future PR

PR #9, opened 2023-09-14
Add reverse iperf testing

SalDaniele:
This commit adds reverse testing. As a side effect it also allows multiple connections be run in a single test.

Additional type safety is added by using dataclass for all instances of json / dict data.


SalDaniele:
@vrindle PTAL

vrindle:
LGTM

PR #8, opened 2023-09-14
[DO NOT MERGE - DRAFT] Add initial revision of hardware offload

vrindle:

PR #10, opened 2023-09-22
Aggregate test results

SalDaniele:
Aggregate logs to single file

Rather than outputing each individual flow result to a separate file,
compile all tests to a single log. Refactor evaluator to accept a single
file.

Add basic outline to allow plugins to output results. Add a temporary
placeholder for measureCpu and measurePower to store results as json.

TODO: Update plugins to sync with test case intervals.
TODO: Update evaluator to evaluate each plugin

### DIFF
--- a/common.py
+++ b/common.py
@@ -1,10 +1,11 @@
 import jinja2
-from dataclasses import dataclass
+from dataclasses import dataclass, asdict, field
 from enum import Enum
-
+from typing import List
 
 FT_BASE_IMG = "quay.io/wizhao/ft-base-image:0.9"
 TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:0.3"
+TFT_TESTS = "tft-tests"
 
 class TestType(Enum):
     IPERF_TCP  = 1
@@ -75,6 +76,27 @@ class IperfOutput():
     tft_metadata: TestMetadata
     command: str
     result: dict
+
+@dataclass
+class PluginOutput():
+    plugin_metadata: dict
+    command: str
+    result: dict
+    name: str
+
+@dataclass
+class TftAggregateOutput():
+    '''Aggregated output of a single tft run. A single run of a trafficFlowTests._run_tests() will
+    pass a reference to an instance of TftAggregateOutput to each task to which the task will append
+    it's respective output. A list of this class will be the expected format of input provided to
+    evaluator.py.
+
+    Attributes:
+        flow_test: an object of type IperfOutput containing the results of a flow test run
+        plugins: a list of objects derivated from type PluginOutput for each optional plugin to append
+        resulting output to.'''
+    flow_test: IperfOutput = None
+    plugins: List[PluginOutput] = field(default_factory=list)
 
 
 def j2_render(in_file_name, out_file_name, kwargs):
--- a/evaluator.py
+++ b/evaluator.py
@@ -4,7 +4,7 @@ import sys
 import yaml
 import json
 from dataclasses import dataclass, asdict
-from common import TestCaseType, TestType, IperfOutput, TestMetadata
+from common import TestCaseType, TestType, IperfOutput, TestMetadata, TftAggregateOutput, TFT_TESTS
 from logger import logger
 from pathlib import Path
 
@@ -54,15 +54,15 @@ class Evaluator():
         self.config = c
         self.test_results = []
 
-    def eval_log(self, log_path: Path):
-
+    def _eval_flow_test(self, run):
         try:
-            with open(log_path, 'r') as file:
-                data = IperfOutput(**json.load(file))
+            run = TftAggregateOutput(**run)
+            data = IperfOutput(**run.flow_test)
             md = TestMetadata(**data.tft_metadata)
         except Exception as e:
-            logger.error(f"Exception: {e}. Malformed log handed to eval_log()")
-            raise Exception(f"eval_log(): error parsing {log_path} for expected fields")
+            logger.error(f"Exception: {e}. Malformed log handed to _eval()")
+            raise Exception(f"_eval(): error parsing data for expected fields")
+
         bitrate_threshold = self.get_threshold(md.test_case_id, md.test_type)
         bitrate_gbps = self.calculate_gbps(data.result, md.test_type)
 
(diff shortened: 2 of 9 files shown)


SalDaniele:
@wizhaoredhat @vrindle PTAL. This consolidates the logs into a single file, and simplifies logging. It also adds a basic placeholder for how plugin's can write their logs to the output.

@vrindle feel free to merge #8  first and I can handle the rebase, but if you are a bit stumped for how to write logs from plugins this might provide some inspiration.

vrindle:
> @wizhaoredhat @vrindle PTAL. This consolidates the logs into a single file, and simplifies logging. It also adds a basic placeholder for how plugin's can write their logs to the output.
> 
> @vrindle feel free to merge #8 first and I can handle the rebase, but if you are a bit stumped for how to write logs from plugins this might provide some inspiration.

No this is helpful. We can merge this first.

common.py
```diff
@@ -76,6 +76,34 @@ class IperfOutput():
...
+    result: dict
+    name: str
+
+@dataclass
+class TftAggregateOutput():
+    iperf_output: IperfOutput = None
+    measureCPU_client_output: PluginOutput = None
+    measureCPU_server_output: PluginOutput = None
+    measurePower_client_output: PluginOutput = None
+    measurePower_server_output: PluginOutput = None
```
vrindle, line 92: All these types should be wrapped in optional since they are not required.

```
from typing import Optional

...

    iperf_output: Optional[IperfOutput] = None

File: evaluator.py [hunk 2]
```diff
@@ -54,15 +54,14 @@ def __init__(self, config_path: str):
...
+    def _eval_flow_test(self, run):
         try:
-            with open(log_path, 'r') as file:
-                data = IperfOutput(**json.load(file))
+            data = IperfOutput(**run["flow_test"])
             md = TestMetadata(**data.tft_metadata)
-        except KeyError as e:
-            logger.error(f"KeyError: {e}. Malformed log handed to eval_log()")
-            raise Exception(f"eval_log(): error parsing {log_path} for expected fields")
+        except Exception as e:
```
vrindle, line 61: can we catch a more specific exception?

measureCpu.py
```diff
@@ -42,7 +45,28 @@ def stop(self):
...
+    #TODO: We are currently only storing the "cpu: all" data from mpstat
+    def generate_output(self, data) -> PluginOutput:
+        return PluginOutput(
+            plugin_metadata={
+                "name": "MeasureCPU",
+                "node_name": self.node_name,
+                "pod_name": self.pod_name,
+            },
+            command=self.cmd,
+            result=data[0],
```
vrindle, line 70: Do we want this entire output or could we parse it to be something more useful?

trafficFlowTests.py
```diff
@@ -85,123 +87,115 @@ def run_tests(self, servers, clients, monitors, duration: int):
...
+        json_out = {"runs": []}
+        for out in self.tft_output:
+            json_out["runs"].append(
+                out.dump_to_dict()
+            )
+        with open(log, "w") as output_file:
+            json.dump(json_out, output_file)
+
+    def evaluate_run_success(self) -> bool:
+        all_passing = True
```
vrindle, line 114: what is this variable for? It doesn't look like it's being used.

evaluator.py
SalDaniele: Yeah, we could add a catch for both type error and key error which might occur like we did before, but I figured it would be better to handle this more generally since we're just logging and passing the exception along.

measureCpu.py
SalDaniele: We definitely want to parse it eventually, however this is an early implementation of measureCPU. Eventually we will need to synchronize the intervals of each measurement with the flow test results and generate some sort of machine readable object. For the time being I opted to just put this as a placeholder for something more sophisticated.

trafficFlowTests.py
SalDaniele: Aahh it is being used, but not meaningfully with the new change. I will remove this thanks

measureCpu.py
```diff
@@ -10,28 +10,31 @@
 import jc
 
 class MeasureCPU(Task):
-    def __init__(self, tc: TestConfig, node_name: str, tenant: bool):
+    def __init__(self, tc: TestConfig, node_name: str, tenant: bool, is_client: bool = False):
```
wizhaoredhat, line 13: Let's start adding some commentary on arguments.  I am not sure if the measure CPU plugin needs to be aware that it is measuring client vs. server. Or we can couple it by providing a reference to the Iperf/Traffic Flow instance it wants to measure. Wdyt?

measurePower.py
```diff
@@ -7,22 +7,24 @@
 import sys
 import yaml
 import json
-import jc
 import re
 import time
 
 class MeasurePower(Task):
-    def __init__(self, tc: TestConfig, node_name: str, tenant: bool):
+    def __init__(self, tc: TestConfig, node_name: str, tenant: bool, is_client: bool = False):
```
wizhaoredhat, line 14: Same comments as I had with measure CPU.

wizhaoredhat (commented):
Overall looks good! Nice work. Just some comments.

However I am not sure how I feel about having "_" before internal functions. Does this add anything or is it a styling preference?

SalDaniele:
> Overall looks good! Nice work. Just some comments.
> 
> However I am not sure how I feel about having "_" before internal functions. Does this add anything or is it a styling preference?

Nope, doesn't do anything, just a styling preference in python to specify internal functions. It is not enforced by the interpreter or anything

common.py
```diff
@@ -76,6 +77,34 @@ class IperfOutput():
...
+
+@dataclass
+class TftAggregateOutput():
+    iperf_output: IperfOutput = None
+    measureCPU_client_output: Optional[PluginOutput] = None
+    measureCPU_server_output: Optional[PluginOutput] = None
+    measurePower_client_output: Optional[PluginOutput] = None
+    measurePower_server_output: Optional[PluginOutput] = None
+
+    def dump_to_dict(self) -> dict:
```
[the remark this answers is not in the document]
SalDaniele: Yeah I was thinking this would be a better way to do this as well. I had a previous iteration that required something a bit more involved, but I think a list of plugins would make more sense. I will add this

evaluator.py
```diff
@@ -54,15 +54,14 @@ def __init__(self, config_path: str):
         self.config = c
         self.test_results = []
 
-    def eval_log(self, log_path: Path):
-
+    def _eval_flow_test(self, run):
         try:
-            with open(log_path, 'r') as file:
-                data = IperfOutput(**json.load(file))
+            data = IperfOutput(**run["flow_test"])
```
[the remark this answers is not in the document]
SalDaniele: Yes that would be better, will add this as well

measureCpu.py
SalDaniele: Yeah this would not be needed by updating the  TftAggregateOuput to handle plugins abstractly.

This was just to ensure a unique field for the output of each of these plugins: i.e. https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/10/files#diff-694b083fdb1c62ebbbbe1e0c047b094b3eb64e1b301494c9d3d0f429c87d4d10R73

measurePower.py
SalDaniele: Removed in recent update

wizhaoredhat (commented):
Very small NIT. Looks good. Good to merge.

PR #11, opened 2023-09-24
Fixed containerfile to add working crictl. Updated TFT tools image to…

vrindle:
… a working image
Current docker file did not have a working crictl command. This commit makes it so that the dockerfile now has an updated crictl. In addition, this also an updated image with the updated crictl.

### DIFF
--- a/common.py
+++ b/common.py
@@ -4,7 +4,7 @@ from enum import Enum
 from typing import List
 
 FT_BASE_IMG = "quay.io/wizhao/ft-base-image:0.9"
-TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:0.3"
+TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:09-27-2023"
 TFT_TESTS = "tft-tests"
 
 class TestType(Enum):
--- a/images/Containerfile
+++ b/images/Containerfile
@@ -1,7 +1,9 @@
 FROM quay.io/centos/centos:stream9
 
-RUN curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable.repo https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/CentOS_8/devel:kubic:libcontainers:stable.repo \
-curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable:cri-o:${VERSION}.repo https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable:cri-o:${VERSION}/CentOS_8/devel:kubic:libcontainers:stable:cri-o:${VERSION}.repo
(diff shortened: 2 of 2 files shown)


common.py
```diff
@@ -3,7 +3,7 @@
 
 
 FT_BASE_IMG = "quay.io/wizhao/ft-base-image:0.9"
-TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:0.3"
+TFT_TOOLS_IMG = "quay.io/vpunj/tft-tools:0.1-x86_64"
```
SalDaniele, line 6: We should probably keep this consistent / ensure it is a multi-arch image

SalDaniele: More specifically, if this is working, the github action in this repo should automatically build the x86/arm images, create a manifest, and push it to `quay.io/wizhao/tft-tools:latest`

https://github.com/wizhaoredhat/ocp-traffic-flow-tests/actions/runs/6200895676

wizhaoredhat: I second this comment. We should have it multi-arch.

images/Containerfile
```diff
@@ -1,8 +1,11 @@
 FROM quay.io/centos/centos:stream9
 
-RUN curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable.repo https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/CentOS_8/devel:kubic:libcontainers:stable.repo \
-curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable:cri-o:${VERSION}.repo https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable:cri-o:${VERSION}/CentOS_8/devel:kubic:libcontainers:stable:cri-o:${VERSION}.repo
+RUN curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable.repo https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/CentOS_8/devel:kubic:libcontainers:stable.repo
```
wizhaoredhat, line 3: Do we need this, when you are extracting a tar to get crictl?

So it is not possible to get crictl from yum? Did your try a different base image (instead of centos:stream9)?

wizhaoredhat (commented):
Perhaps we need to investigate different UBIs. Please also consider multiarch.

common.py
vrindle: sounds good. I'll do that instead.

wizhaoredhat:
LGTM

common.py
```diff
@@ -4,7 +4,7 @@
 from typing import List
 
 FT_BASE_IMG = "quay.io/wizhao/ft-base-image:0.9"
-TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:0.3"
+TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:09-27-2023"
```
wizhaoredhat, line 7: @vrindle Can you try quay.io/wizhao/tft-tools:latest ?

PR #12, opened 2023-09-24
Adds getting VF Rep component.

vrindle:
Add first component of HWOL plugin. This gets the VF rep using the tools pod.


SalDaniele:
I'm not sure if this needs to be its own task that would run in the way the other tasks do (i.e. init, set up pods, run, wait on stop for run thread to finish). Seems like this could just be a function on the task that handles validating HWOL.

getVfRep.py
```diff
@@ -0,0 +1,49 @@
...
+from thread import ReturnValueThread
+from task import Task
+from host import Result
+from typing import Optional
+import sys
+import yaml
+import json
+import jc
+
+class GetVfRep(Task):
```
wizhaoredhat, line 13: I think this class name needs a bit of work. We aren't trying to GetVfRep. Getting the VF rep is a requirement, but what we are really want this plugin to do is to get the VF rep stats and traces.

getVfRep.py
```diff
@@ -0,0 +1,49 @@
...
+    def run(self, duration: int):
+        self.exec_thread = ReturnValueThread(target=self.run_st)
+        self.exec_thread.start()
+
+    def stop(self):
+        logger.info(f"Stopping Get Vf Rep execution on {self.pod_name}")
+        r = self.exec_thread.join()
+        if r.returncode != 0:
+            logger.info(r)
+        ethtool_output = r.out
```
wizhaoredhat, line 40: This is unused?

getVfRep.py
```diff
@@ -0,0 +1,49 @@
...
+        data = json.loads(r.out)
+        self.vf_rep= data["containers"][0]["podSandboxId"][:15]
```
wizhaoredhat, line 42: Getting the VF rep name is the first part. But we need to run at the very least ethtool -S on it.

wizhaoredhat (commented):
I think we need to do a little more for this PR to be a working plugin. At the very least ethtool stats.

vrindle:
> I think we need to do a little more for this PR to be a working plugin. At the very least ethtool stats.

I understand that this is not a full working plugin, but I wanted to make it easier for you to review which is why I'm splitting this into smaller components. If you'd like in ethtool stat and tcp dump and all of that stuff as 1 big pr.

wizhaoredhat:
> > I think we need to do a little more for this PR to be a working plugin. At the very least ethtool stats.
> 
> I understand that this is not a full working plugin, but I wanted to make it easier for you to review which is why I'm splitting this into smaller components. If you'd like in ethtool stat and tcp dump and all of that stuff as 1 big pr.

It was discussed that getting the VF Rep shouldn't be the main task that your plugin is running. Getting the VFRep should be a function. The first iteration could be simply running ethtool -S periodically (without tcpdump)

getVfRep.py
SalDaniele: Agreed with William's point

task.py
```diff
@@ -75,7 +76,7 @@ def create_node_port_service(self, nodeport: int) -> str:
         logger.info(f"Creating Node Port Service {out_file_yaml}")
         r = self.run_oc(f"apply -f {out_file_yaml}")
         if r.returncode != 0:
-            if "already exists" not in r.err:
+            if "already exists" not in r.err and self.reuse_node:
```
SalDaniele, line 79: Should this be reuse pod? Maybe just override this function in your task rather than needing to add another switch to the generic Task ABC

SalDaniele, line 79: Why do we need this switch? We aren't creating a new node port service from your task, correct?

getVfRep.py
```diff
@@ -0,0 +1,98 @@
...
+        
+
+    def extract_vf_rep(self) -> str:
+        get_vf_rep_cmd = f'exec -n default {self.pod_name} -- /bin/sh -c "crictl --runtime-endpoint=/host/run/crio/crio.sock ps -a --name={self.iperf_pod_name} -o json "'
+        vf_rep_json = self.run_oc(get_vf_rep_cmd).out
+        data = json.loads(vf_rep_json)
+        return data["containers"][0]["podSandboxId"][:15]
+
+    def run_ethtool_cmd(self, vf_rep: str) -> (int, int):
+        self.ethtool_cmd = f'exec -n default {self.pod_name} -- /bin/sh -c "ethtool -S {vf_rep}"'
```
SalDaniele, line 37: We should check to make sure this does not fail

getVfRep.py
```diff
@@ -0,0 +1,98 @@
...
+import yaml
+import json
+import jc
+
+class GetVfRep(Task):
+    def __init__(self, tft: TestConfig, iperf_instance: Union[IperfServer, IperfClient], tenant: bool):
+        super().__init__(tft, 0, iperf_instance.node_name, tenant, True)
+
+        self.in_file_template = "./manifests/tools-pod.yaml.j2"
+        self.out_file_yaml = f"./manifests/yamls/tools-pod-{self.node_name}-get-vf-rep.yaml"
```
SalDaniele, line 19: Same comment about the naming here, this pod is more general than just getting the vf rep

getVfRep.py
```diff
@@ -0,0 +1,98 @@
...
+        tx_packet_start = self._output_ethtool.result["tx_packet_start"]
+        rx_packet_end = self._output_ethtool.result["rx_packet_end"]
+        tx_packet_end = self._output_ethtool.result["tx_packet_end"]
+
+        logger.info("rx_packet_start: %d\n tx_packet_start: %d\n rx_packet_end: %d\n tx_packet_end: %d\n" % (rx_packet_start,tx_packet_start,rx_packet_end,tx_packet_end))
+
+    def generate_output_ethtool(self, data, cmd: str) -> PluginOutput:
+        return PluginOutput(
+            plugin_metadata={
+                "name": "GetEthtoolStats",
```
SalDaniele, line 90: This could also be a bit more specific. We should use the same plugin name everywhere i.e. pod name, plugin name provided by the user, class name etc. Similar to measureCpu/Power

task.py
```diff
@@ -11,13 +11,14 @@
 import host
 
 class Task(ABC):
-    def __init__(self, tft: TestConfig, index: int, node_name: str, tenant: bool):
+    def __init__(self, tft: TestConfig, index: int, node_name: str, tenant: bool, reuse_node: bool = False):
```
SalDaniele, line 14: Can we get rid of this?

getVfRep.py
```diff
@@ -0,0 +1,98 @@
...
+        self.template_args["test_image"] = TFT_TOOLS_IMG
+
+        self.pod_name = self.template_args["pod_name"]
+        self.iperf_pod_name = iperf_instance.template_args["pod_name"]
+
+        j2_render(self.in_file_template, self.out_file_yaml, self.template_args)
+        logger.info(f"Generated Server Pod Yaml {self.out_file_yaml}")
+        
+
+    def extract_vf_rep(self) -> str:
```
SalDaniele, line 30: Would be good to add some debug / info logging here

getVfRep.py
```diff
@@ -0,0 +1,98 @@
...
+        get_vf_rep_cmd = f'exec -n default {self.pod_name} -- /bin/sh -c "crictl --runtime-endpoint=/host/run/crio/crio.sock ps -a --name={self.iperf_pod_name} -o json "'
+        vf_rep_json = self.run_oc(get_vf_rep_cmd).out
+        data = json.loads(vf_rep_json)
+        return data["containers"][0]["podSandboxId"][:15]
+
+    def run_ethtool_cmd(self, vf_rep: str) -> (int, int):
+        self.ethtool_cmd = f'exec -n default {self.pod_name} -- /bin/sh -c "ethtool -S {vf_rep}"'
+        ethtool_output = self.run_oc(self.ethtool_cmd).out
+        #print("DEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEBAAAAAAAAAAAAAG %s" % ethtool_output)
```
wizhaoredhat, line 39: Left over debug?

getVfRep.py
```diff
@@ -0,0 +1,98 @@
...
+        rxpacket = self.parse_out_packet(ethtool_output, "rx_packet")
```
wizhaoredhat, line 40: As discussed please add a delay/sleep here for now.

Please look into semaphores and barriers for synchronization techniques between different running processes.

getVfRep.py
vrindle: lol this is my spelling of debug ;P

task.py
vrindle: I think that because we are passing in the node from the iperf class the node port service is already in service, so this a way of handling that instance. The reason why we are not overriding this function it is unneccessary code duplication for one conditional.

getVfRep.py
vrindle: I am planning on creating at least 2 plugin outputs for this class (GetEthtoolStats as well as TCP dump stats). This class is really going to encapsulate all of HWOL and generate more than 1 plugin output so we can't have naming consistency.

getVfRep.py
vrindle: In that case we should change up the structure of the other plugins because they use a similar naming mechanism: https://github.com/vrindle/ocp-traffic-flow-tests/blob/d611ea6823939dff8c08cf99c9f123778aa7b24f/measureCpu.py#L17C77-L17C93

task.py
vrindle: Tell me what you think about this: https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/12#discussion_r1345082034

getVfRep.py
vrindle: I can add a delay/sleep but just to clarify all these commands are currently being run synchronously on a single thread. So because this is running on a single thread I'm curious as to how we will use semaphores?

common.py
```diff
@@ -4,7 +4,7 @@
 from typing import List
 
 FT_BASE_IMG = "quay.io/wizhao/ft-base-image:0.9"
-TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:09-27-2023"
+TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:0.3"
```
SalDaniele, line 7: If the github actions automation is working as expected, we should be using the latest image rather than a hardcoded version

config.yaml
```diff
@@ -12,13 +12,13 @@ tft:
...
             sriov: "true"
         client:
-          - name: "worker-230"
+          - name: "worker-236"
             sriov: "true"
         plugins:
           - name: capture_flows
-          - name: measure_cpu
-          - name: measure_power
+          - name: get_vf_rep
```
SalDaniele, line 23: Let's leave this as an example config file for users to reference with all the plugin options specified. If you only want to test w/ a subset, I would use an untracked personal config file

getVfRep.py
SalDaniele: To clarify, the plugin "measureCpu" measures the CPU usage using mpstat, or the "measurePower" plugin measures the power usage etc.

This comment is just to say that we should determine a name for what this plugin does "i.e. validate-vf-rep-offload or something" and use this consistently for the task name, string provided by user to invoke the plugin, yaml files required to run the plugin etc.

SalDaniele: The name of this plugin implies this just gets the vf rep name. We are doing more than that

getVfRep.py
SalDaniele: There are multiple threads running at once i.e. the actual iperf flow tests, the other plugins. When we run task.start() on each task, these tasks will run concurrently.

For now we can just add a sleep so that your plugin will run roughly the duration of the iperf flow test. However, a better way to do this that we will need to figure out for all plugins is to synchronize collecting plugin data with the internals of the iperf test. This would require some sort of semaphore or barrier to ensure that T1 in one task corresponds to T1 in every other tasks.

task.py
SalDaniele: What is this switch doing? If the pod already exists, we will get an error, however we can ignore this since the pod we want is running.

I don't think we are setting this switch anywhere? I'm not sure what this is trying to accomplish, if we remove this now, nothing would change, no?

SalDaniele:
Most inline comments are more on the nit-pick side, the main issue that needs resolving is getting the plugin to a state where it can be run, and accurately generates output showing the ethtool stats on the vf rep before and after the proper interval.

test.py
```diff
@@ -0,0 +1,18 @@
+output = """NIC statistics:
```
SalDaniele, line 1: I assume this needs to be removed?

task.py
vrindle: The tests don't work without this switch, hence why it was added in.

SalDaniele: Could you elaborate why? What I don't understand is we don't call this function from this new plugin / any task other than the iperf_server task, and this currently works without these changes.

With this change if any error occurs when the iperf server is creating the node port service, it will be ignored. Is this the desired behavior?

test.py
vrindle: yes. Accidentally committed.

task.py
vrindle: We pass in the node name from the iperf class and we reuse the node and it seems to hit the code path again and exit the test prematurely.

```With this change if any error occurs when the iperf server is creating the node port service, it will be ignored. Is this the desired behavior?```

This is not the case. We only set reuse_node to true when using the VF Rep class. That's the only instance it will forgo the check. Otherwise, the check is still valid.

Comment (SalDaniele) [reply task.py:79]: Lets say we are running the tests without using your plugin.

If "already exists" is not in r.err, that means a valid error has occurred, as we need to exit.

Default for self.reuse_node = False, so this will be the case when we create the iperf server instance.

Therefore, if we start an iperf server and create a node port service, and a failure occurs, it will be `if <True> and <False>` and not exit as it should. Am I thinking about this correctly?

Comment (SalDaniele) [reply task.py:79]: Why do we not need this check on create_cluster_ip_server()? And more importantly, are we creating a second iperf server somehow? We only call these functions from iperf_server.init()

Comment (wizhaoredhat) [reply task.py:79]: I don't understand the need for this flag... If you can provide us with a situation where it would be needed then we can better understand.

Comment (vrindle) [reply task.py:79]: I apologize. I hit a concurrency bug earlier and believed I needed this flag. You are correct. We do not need this flag. Removing

File: validateOffload.py [hunk 15]
```diff
@@ -0,0 +1,123 @@
...
+
+    def parse_out_packet(self, output: str, prefix: str) -> Optional[int]:
+        for line in output.splitlines():
+            stripped_line = line.strip()
+            if stripped_line.startswith(prefix):
+                return int(stripped_line.split(":")[1])
+
+        return None
+    
+    def run_st(self) -> (int, int, int, int):
```
SalDaniele, line 73: Can these args be cleaned up?

SalDaniele:
LGTM

SalDaniele:
Tested and it looks like it is working as expected

SalDaniele (approved):

trafficFlowTests.py
```diff
@@ -166,6 +173,10 @@ def _run(self, connections: dict, test_type: TestType, test_id: int, index: int,
                     self._enable_measure_cpu_plugin(monitors, node_server_name, node_client_name, True)
                 if plugins['name'] == "measure_power":
                     self._enable_measure_power_plugin(monitors, node_server_name, node_client_name, True)
+                if plugins['name'] == "validate_offload":
+                    iperf_server = servers[-1]
```
wizhaoredhat, line 177: I think we need to put a TODO here. You are taking the last server at the end of the queue. But we should be running this on each individual server + client pairs.

wizhaoredhat:
Overall LGTM. Just the minor comment and this can be merged.

PR #14, opened 2024-03-22
Lint

Zinkelburger:
My changes to lint, refactor, and debug

I have confirmed it has an exact match to `main` for iperf.py test logs, and the evaluator results. 

I'm going through Thomas' comments, and will cherry pick out sections to individual PRs to make it more manageable to review.


wizhaoredhat:
Let me know once this is good to merge.

arguments.py
```diff
@@ -1,31 +1,52 @@
 import os
-from logger import logger, configure_logger
+from logger import configure_logger
```
thom311, line 1: It would seem nicer to separate the commit in an automated part vs. manual modification.

One might assume, that the tool behaves correctly (or at least, it does what it does). So reviewing the manual changes would be especially interesting. But it's hard to see which changes were done by hand.

Or were the unnecessary imports removed by a tool? How?

In any case, the command line you used to make this commit would be nice too. I presume it's just `black .`?

Zinkelburger: Yes. I'll put that in a separate PR

common.py
```diff
@@ -1,7 +1,7 @@
 import jinja2
 from dataclasses import dataclass, field
 from enum import Enum
-from typing import List
+from typing import List, Optional, Any, Dict, List, Union, Type, TypeVar, cast
```
thom311, line 4: If this depends on Python 3.9+, then you don't need the deprecated `typing.{List,Dict,Tuple}` type but can use `{list,dict,tuple}` directly. See https://docs.python.org/3/library/typing.html#typing.List

Depending on Python 3.9+ seems reasonable since,
- it's in RHEL9
- CDA also depends on Python 3.11+

btw, `List` is imported twice. I would think that a linter complains about that too...

thom311: (it's fine, this can be changed in a follow-up commit too. It it causes extra work to rebase)

evaluator.py
```diff
@@ -62,14 +63,17 @@ def __init__(self, config_path: str):
...
+        }
         self.test_results: List[TestResult] = []
 
-    def _eval_flow_test(self, run):
+    def _eval_flow_test(self, run) -> None:
         try:
             run = TftAggregateOutput(**run)
             data = IperfOutput(**run.flow_test)
-            md = TestMetadata(**data.tft_metadata)
+            md = from_dict(TestMetadata(**data.tft_metadata))
```
thom311: this call lacks a `data` argument.

However, a followup commit refactors this again.

testSettings.py
```diff
@@ -106,33 +106,38 @@ def get_test_metadata(self) -> TestMetadata:
...
-    def server_test_to_pod_type(self, test_id: int, cfg_pod_type: str) -> PodType:
-        if test_id in (3, 4, 7, 8, 19, 20, 23, 24):
+    def server_test_to_pod_type(
+        self, test_id: TestCaseType, cfg_pod_type: PodType
+    ) -> PodType:
+        if test_id.value in (3, 4, 7, 8, 19, 20, 23, 24):
             return PodType.HOSTBACKED
 
-        if cfg_pod_type == "sriov":
+        if cfg_pod_type.name == "SRIOV":
```
thom311, line 131: `if cfg_pod_type == PodType.SRIOV:`

(below too)

common.py
```diff
@@ -71,18 +87,20 @@ class PodInfo:
 
 @dataclass
 class TestMetadata:
-    test_case_id: TestCaseType
-    test_type: TestType
     reverse: bool
-    server: PodInfo
-    client: PodInfo
+    test_case_id: TestCaseType = field(default_factory=enum_factory(TestCaseType))
```
thom311: does this actually work? `default_factory` should be a callable without any argument. But `enum_factory()` returns a callable that takes a `value:Any` argument.

common.py
```diff
@@ -8,6 +8,22 @@
 TFT_TESTS = "tft-tests"
 
 
+def enum_factory(enum_type: Type[Enum]):
```
thom311: this lacks a return type annotation.

common.py
```diff
@@ -8,6 +8,22 @@
...
+    def factory(value: Any) -> Enum:
```
thom311: granted, the function handles `Any` type (to some definition of the work "handle"). But in practice, it only accepts strings or the enum type.

Wouldn't it be better, to tell the type checker about that, and return a `value: str | T` value (where `T` is the Enum type)? Because passing anything else is not going to work, and likely a bug we want to be told by the type checker.

common.py
```diff
@@ -71,18 +87,20 @@ class PodInfo:
...
+    client: PodInfo = field(default_factory=lambda: from_dict(PodInfo, {}))
 
 
 @dataclass
 class IperfOutput:
-    tft_metadata: TestMetadata
     command: str
     result: dict
+    tft_metadata: TestMetadata = field(
+        default_factory=lambda: from_dict(TestMetadata, {})
```
thom311: why not `lambda: TestMetaData()` ?

common.py
```diff
@@ -1,27 +1,26 @@
...
-                return member
-            else:
-                raise ValueError(f"{value} is not a valid member of {enum_type}")
-        else:
-            raise ValueError(f"Cannot convert {value} to {enum_type}")
+E = TypeVar("E", bound=Enum)
+
 
-    return factory
+def enum_convert(enum_type: Type[E], value: Any) -> E:
```
thom311, line 14: ah, I commented about the `enum_factory` already. Maybe it would be good to rebase the branch to not first introduce code that gets modified/fixed by a followup value.

Still, I think it should be `value: str | E`.

common.py
```diff
@@ -173,12 +190,6 @@ def from_dict(cls: Type[T], data: Dict[str, Any]) -> T:
         field_type = field.type
```
thom311, line 197: as this function only works with data classes, it would have been a bit nicer to call it `dataclass_from_dict()`. But I see there are a lot of usages already. It's cumbersome to fix now... as you prefer.

common.py
```diff
@@ -173,12 +190,6 @@ def from_dict(cls: Type[T], data: Dict[str, Any]) -> T:
...
-        elif (
-            isinstance(field.type, type)
-            and issubclass(field.type, Enum)
-            and field_name in data
-        ):
-            field_values[field_name] = enum_factory(field.type)(data[field_name])
         elif field_name in data:
             field_values[field_name] = data[field_name]
-    return cls(**field_values)
+    return cast(T, cls(**field_values))
```
thom311, line 202: this function is was introduced by the branch, and modified twice afterwards (on the same branch). Can this part of the branch be rebased?

measurePower.py
```diff
@@ -62,21 +62,25 @@ def stat(self, cmd: str, duration: int):
...
-            logger.error(r)
-        self._output = self.generate_output(data=r.out)
+        self.exec_thread.join()
+        if self.exec_thread.result is not None:
+            r = self.exec_thread.result
+            if r.returncode != 0:
+                logger.error(r)
+            self._output = self.generate_output(data=r.out)
+        else:
+            logger.error("Thread did not return a result")
```
thom311: `self._output` should be set to an empty `PluginOutput`.

You already handle the error (by logging a message). If you leave `self._output` unset, a crash will later happen, when trying to access the output. The error is handled, and there must be some usable output set, so that followup code can do something with that.

The function later is refactored, but it still applies then.

thom311: probably, just call `self.generate_output(data=None)` and those functions need to return something usable. 

`generate_output()` could also return `None` (an `Optional[common.BaseOutput]`). But then every user has to handle None specially. Better to return some output with no data.

iperf.py
```diff
@@ -84,6 +84,19 @@ def server(self, cmd: str) -> Result:  # type: ignore
...
+        if self.exec_thread.result is not None:
+            r = self.exec_thread.result
+            if r.returncode != 0:
+                logger.error(
+                    f"Error occured while stopping Iperf server: errcode: {r.returncode} err {r.err}"
+                )
+            logger.debug(f"IperfServer.stop(): {r.out}")
+        else:
+            logger.error("Thread did not return a result")
+
```
thom311, line 100: why is the stop method returned? The commit message should explain.

But better: rebase to not drop it in the first place.

iperf.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, tc: TestConfig, ts: TestSettings):
 
         if self.exec_persistent:
             self.template_args["command"] = IPERF_EXE
-            self.template_args["args"] = ["-s", "-p", f"{self.port}"]
+            self.template_args["args"] = ["-s", "-p", f"{self.port}"]  # type: ignore
```
thom311, line 60: this seems wrong to me.

Why is this done?

Maybe the type should be `dict[str, str|list[str]]` or `Mapping[str, str | list[str]]`.

common.py
```diff
@@ -19,6 +19,11 @@ def enum_convert(enum_type: Type[E], value: Any) -> E:
             return enum_type[value]
         except KeyError:
             raise ValueError(f"Cannot convert {value} to {enum_type}")
+    elif isinstance(value, int):
+        try:
+            return enum_type(value)
+        except ValueError:
+            raise ValueError(f"Cannot convert {value} to {enum_type}")
```
thom311, line 26: this part should move to the patch that first introduced `enum_convert()`. The change is also not specifically needed by this patch.

thom311 (commented):
The overall direction of the branch is IMO good.

I was thinking, it may not be worth the effort to make every commit self-contained and beautiful. E.g. rebase changes around, etc. If you think it's not the best way to spend your time, I think it can be merged without.

iperf.py
Zinkelburger: I tried a `dict[str, str|list[str]` at one point, but it made many more problems than it solved. 
Since args is the only thing that's a `list[str]`, I thought I defined at one point to use `self.args` instead of `self.template_args["args"]`. Must have gotten lost.

iperf.py
Zinkelburger: I am making this into its own PR, will revert the `stop` related commits from this branch

common.py
Zinkelburger: It doesn't work, I remove it later

common.py
Zinkelburger: Good suggestion, I added that

common.py
Zinkelburger: Good point. I'm not sure why I did that.
I removed this and refactored the code

measurePower.py
Zinkelburger: That is a good point. I added this line to the `stop` refactor: [self._output = common.BaseOutput("", {})](https://github.com/wizhaoredhat/ocp-traffic-flow-tests/blob/main/task.py#L125)

SalDaniele:
Closing as these changes are now submitted in #24

PR #16, opened 2024-04-17
Lint with black, remove unnecessary imports

Zinkelburger:
Beginning the process of separating out the linting PR.

This commit is simply the result of running `black .`  and manually removing unused imports.


SalDaniele:
lgtm

PR #17, opened 2024-04-17
Refactor task.py abstract `stop`

Zinkelburger:
`stop` is used almost the same way by all of Task's derived classes. The derived classes similarly all use a `generate_output` function.

1. Make `task.py`'s stop not abstract, provide an implementation to join the threads, set `self._output` to `self.generate_output`
2. Create a `common.BaseOutput` class to abstract common functionality of `IperfOutput` and `PluginOutput`
3. Make a `self.generate_output` abstract function, to return a `common.BaseOutput`

- All derived classes now use Task's `stop` implementation
- ValidateOffload was fixed to correctly count packets. The previous implementation always returned the default value of "None"

This is one of the most potentially breaking changes, and I have several commits to change `stop` and `generate_output`, so I squashed them into this PR
I'll test that its non-breaking before I mark it as ready for review.

### DIFF
--- a/iperf.py
+++ b/iperf.py
@@ -159,6 +159,9 @@ class IperfClient(Task):
 
     def output(self, out: common.TftAggregateOutput):
         # Return machine-readable output to top level
+        assert isinstance(
+            self._output, IperfOutput
+        ), f"Expected variable to be of type IperfOutput, got {type(self._output)} instead."
         out.flow_test = self._output
 
         # Print summary to console logs
--- a/iperf.py
+++ b/iperf.py
@@ -4,7 +4,7 @@ from logger import logger
 from testConfig import TestConfig
 from thread import ReturnValueThread
 from task import Task
-from host import Result
+from common import Result
 from testSettings import TestSettings
 import json
 
@@ -74,7 +74,7 @@ class IperfServer(Task):
 
         logger.info(f"Running {cmd}")
 
-        def server(self, cmd: str):
+        def server(self, cmd: str) -> Result:
             if self.connection_mode == ConnectionMode.EXTERNAL_IP:
                 return self.lh.run(cmd)
             elif self.exec_persistent:
@@ -135,8 +135,8 @@ class IperfClient(Task):
         common.j2_render(self.in_file_template, self.out_file_yaml, self.template_args)
         logger.info(f"Generated Client Pod Yaml {self.out_file_yaml}")
 
-    def run(self, duration: int):
-        def client(self, cmd: str):
+    def run(self, duration: int) -> None:
+        def client(self, cmd: str) -> Result:
             return self.run_oc(cmd)
 
         server_ip = self.get_target_ip()
--- a/iperf.py
+++ b/iperf.py
@@ -84,20 +84,14 @@ class IperfServer(Task):
         self.exec_thread = ReturnValueThread(target=server, args=(self, cmd))
         self.exec_thread.start()
 
-    def run(self, duration: int):
+    def run(self, duration: int) -> None:
         pass
 
-    def stop(self):
-        logger.info(f"Stopping execution on {self.pod_name}")
-        r = self.exec_thread.join()
-        if r.returncode != 0:
-            logger.error(
-                f"Error occured while stopping Iperf server: errcode: {r.returncode} err {r.err}"
-            )
-        logger.debug(f"IperfServer.stop(): {r.out}")
-
-    def output(self, out: common.TftAggregateOutput):
+    def output(self, out: common.TftAggregateOutput) -> None:
         pass
+
+    def generate_output(self, data: str) -> common.BaseOutput:
+        return common.BaseOutput("", {})
 
 
 class IperfClient(Task):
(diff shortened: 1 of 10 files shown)


trafficFlowTests.py
```diff
@@ -109,13 +109,13 @@ def _run_tests(
         for tasks in servers + clients + monitors:
             tasks.setup()
 
-        for tasks in servers + clients + monitors:
+        for tasks in clients + monitors:
             tasks.run(duration)
```
thom311, line 113: This change seems wrong. Tasks should be treated all the same, and if a certain task's `run()` has nothing to do, it should just do nothing. But the callers shouldn't know that, they would still call `run()`.

Granted, you also replaced `pass` in
```python
    def run(self, duration: int) -> None:
        raise NotImplementedError("run() not implemented in IperfServer")
```
But why? If there is nothing to do for that type, then it should just silently do nothing and not raise an exception.

Zinkelburger: That makes sense. I thought we wouldn't want to call empty functions, but it does break the pattern. I'll change it back.

iperf.py
```diff
@@ -84,20 +84,27 @@ def server(self, cmd: str):
...
+                logger.error(
+                    f"Error occured while stopping Iperf server: errcode: {r.returncode} err {r.err}"
+                )
+            logger.debug(f"IperfServer.stop(): {r.out}")
+        else:
+            logger.error("Thread did not return a result")
 
-    def output(self, out: common.TftAggregateOutput):
-        pass
+    def run(self, duration: int) -> None:
```
wizhaoredhat, line 100: We are using the abstract class and polymorphism here. This type of software design makes it such that as long as the base class has an implementation it should be able to call it. The advantage is to make it such that we can lump things together and run them.

Zinkelburger: That makes sense. I reverted those changes removing methods from `IperfServer`.

validateOffload.py
```diff
@@ -71,77 +70,81 @@ def run_ethtool_cmd(self, vf_rep: str) -> (int, int):
...
-                return int(stripped_line.split(":")[1])
-
-        return None
+            if prefix in stripped_line and packet_suffix in stripped_line:
+                packet_count = int(stripped_line.split(":")[1].strip())
+                total_packets += packet_count
+                
+        return total_packets
 
-    def run_st(self) -> RxTxData:
```
SalDaniele, line 89: If we are no longer using this class, we should remove it in common.py

validateOffload.py
```diff
@@ -71,77 +70,81 @@ def run_ethtool_cmd(self, vf_rep: str) -> (int, int):
...
+    def run(self, duration: int) -> None:
         self.exec_thread = ReturnValueThread(target=self.run_st)
         self._duration = int(duration)
         self.exec_thread.start()
 
-    def stop(self):
-        logger.info(f"Stopping Get Vf Rep execution on {self.pod_name}")
-        r = self.exec_thread.join()
-
-        if self.iperf_pod_type == PodType.HOSTBACKED:
```
SalDaniele, line 111: We still need to handle this case, maybe need to overload stop for this plugin

SalDaniele: Lets handle this in generate output

validateOffload.py
```diff
@@ -71,77 +70,81 @@ def run_ethtool_cmd(self, vf_rep: str) -> (int, int):
...
+            if prefix in stripped_line and packet_suffix in stripped_line:
+                packet_count = int(stripped_line.split(":")[1].strip())
+                total_packets += packet_count
+                
+        return total_packets
 
-    def run_st(self) -> RxTxData:
+    def run_st(self) -> Result:
         vf_rep = self.extract_vf_rep()
-        (rxpacket_start, txpacket_start) = self.run_ethtool_cmd(vf_rep)
```
SalDaniele, line 91: I feel like this should return a Result. If it fails, we bail early

validateOffload.py
```diff
@@ -71,77 +70,81 @@ def run_ethtool_cmd(self, vf_rep: str) -> (int, int):
...
-        return RxTxData(
-            rx_start=rxpacket_start,
-            tx_start=txpacket_start,
-            rx_end=rxpacket_end,
-            tx_end=txpacket_end,
+        r1 = self.run_ethtool_cmd(vf_rep)
+        time.sleep(self._duration)
+        r2 = self.run_ethtool_cmd(vf_rep)
+
+        combined_out = f"{r1.out}--DELIMIT--{r2.out}"
```
SalDaniele, line 98: This is pretty gross, but I see why you are doing this...I am ok with it for now

SalDaniele: Maybe worth it to keep the design:

run runs the command and generates a result
stop generates the output from the raw result.out

SalDaniele:
Small nits, but LGTM

validateOffload.py
Zinkelburger: Do we need to handle it? The way the code currently handles it is with 
```python
    def extract_vf_rep(self) -> str:
        if self.iperf_pod_type == PodType.HOSTBACKED:
            logger.info(f"The VF representor is: ovn-k8s-mp0_0")
            return "ovn-k8s-mp0_0"
```

wizhaoredhat:
LGTM

PR #18, opened 2024-04-18
Fix for "External" tests (25 & 26)

Zinkelburger:
- change from `podman stop` to `podman rm --force` in traffic flow tests cleanup
- change `-itd` to `-it` so the process is only joinable when the container is done
- add `--one-off` so iperf dies after a connection, causing the container to die
- have `get_podman_ip` try to get the server's ip for 10 seconds


trafficFlowTests.py
```diff
@@ -70,7 +70,7 @@ def _cleanup_previous_testspace(self, namespace: str):
         logger.info(
             f"Cleaning external containers {iperf.EXTERNAL_IPERF3_SERVER} (if present)"
         )
-        cmd = f"podman stop {iperf.EXTERNAL_IPERF3_SERVER}"
+        cmd = f"podman rm --force {iperf.EXTERNAL_IPERF3_SERVER}"
```
wizhaoredhat, line 73: Why do we need --force?

Zinkelburger: It stops the container and removes it, and it is shorter than having `stop` and `rm` separately.
Previously there was an error when trying to create the `EXTERNAL_IPERF3_SERVER pod` because the old pod was only stopped and not removed.

wizhaoredhat: I think force has the unintended side effect that it removes it immediately or "asap". Which means some objects or references aren't cleaned up properly. What do you think about introducing a wait after calling stop to make sure everything is actually gone before proceeding?

Zinkelburger: It is a good point. I was not aware of side affects from `--force`.
I may have [introduced a bug into CDA](https://github.com/bn222/cluster-deployment-automation/pull/141) as a result? I thought it guarantees that the resource was stopped and deleted.

Maybe the best compromise is to do a `stop` and then an `rm --force`, so the container has the chance to gracefully shutdown if it is able to.

Zinkelburger: Stop can also potentially hang, maybe we would need to do stop with a timeout?

wizhaoredhat: I think a timeout would work. Hmm maybe there is a bug but I don't think that code executes all that often.

Zinkelburger: I added a `--time` flag for stop and rm. First it tries to stop the container, then it'll send a SIGKILL

wizhaoredhat: ok, looks good

wizhaoredhat:
LGTM

PR #19, opened 2024-04-19
Add an undo_target function to return value thread

Zinkelburger:
.join() has no way to actually kill the containers it creates. So let `ReturnValueThread` take a `cleanup_action` to undo the `target` action.

- gives threads the possibility to clean themselves up, in case "join" doesn't return
- pass a timeout to `stop`. After this timeout, join will attempt to kill the thread. If it is not able to, the thread is orphaned
- have IperfServer use the new `cleanup_action` parameter

Depends on #17 to add `Result` in common.py, and the new `stop` implementation


wizhaoredhat:
LGTM, PTAL @SalDaniele

SalDaniele:
LGTM, this is a bit redundant since we are already cleaning up all the pods after each test, but I like having a way in the future to handle threads that run beyond the expected duration.

PR #21, opened 2024-04-22
Synchronize threads

### LINKED ISSUE #20: Threads use "duration" to synchronize instead of Events
Threads currently synchronize their readings with a "duration".
Instead, all of the threads should be able to signal that they are ready. Maybe we can even ditch the duration all together and have the client also signal when it ends. 

I put this as an issue because I am leaving this team soon, and this may be a better task for someone more familiar with multithreading in python.

This first came to my attention with ValidateOffload. It
1. Runs a command to get the initial state (no guarantee on when this is happening)
2. [Sleeps for duration](https://github.com/wizhaoredhat/ocp-traffic-flow-tests/blob/main/validateOffload.py#L92)
3. Runs the same command again to get the final state (no guarantee on when this is happening)

One solution could be to use a "Barrier" <https://docs.python.org/3/library/threading.html>
We can make a Barrier of size `len(montiors+servers+clients)`. All threads are blocked until each one calls `barrier.wait`, and they'll start simultaneously.

Zinkelburger:
- Only commit on this branch is [Synchronize threads with SyncManager singleton](https://github.com/wizhaoredhat/ocp-traffic-flow-tests/commit/113d449d855c6c279a4c7ac81de397df23d34d60)
- Rebased on my other changes to Thread & Task to reduce conflicts with them

Resolve #20 

Use barriers to have all of the `ReturnValueThreads` start simultaneously.

Use events for the client to signal when its done, instead of relying on a duration.

### DIFF
--- a/measurePower.py
+++ b/measurePower.py
@@ -7,6 +7,7 @@ from common import Result
 import re
 import time
 import json
+from syncManager import SyncManager
 
 
 class MeasurePower(Task):
@@ -37,29 +38,25 @@ class MeasurePower(Task):
             logger.error(f"Could not find Instantaneous power reading: {e}.")
             return 0
 
-        def stat(self, cmd: str, duration: int) -> Result:
-            end_time = time.time() + float(duration)
+        def stat(self, cmd: str) -> Result:
+            SyncManager.wait_on_barrier()
             total_pwr = 0
             iteration = 0
-            while True:
+            while SyncManager.client_not_finished():
                 r = self.run_oc(cmd)
                 if r.returncode != 0:
                     logger.error(f"Failed to get power {cmd}: {r}")
                 pwr = extract(r)
                 total_pwr += pwr
                 iteration += 1
-                # FIXME: Hardcode interval for now
-                time.sleep(2)
-                if time.time() > end_time:
-                    break
+                time.sleep(0.2)
+
             r = Result(f"{total_pwr/iteration}", "", 0)
             return r
 
         # 1 report at intervals defined by the duration in seconds.
-        self.cmd = f"exec {self.pod_name} -- ipmitool dcmi power reading"
-        self.exec_thread = ReturnValueThread(
-            target=stat, args=(self, self.cmd, duration)
-        )
+        self.cmd = f"exec -t {self.pod_name} -- ipmitool dcmi power reading"
+        self.exec_thread = ReturnValueThread(target=stat, args=(self, self.cmd))
         self.exec_thread.start()
         logger.info(f"Running {self.cmd}")
 
--- a/syncManager.py
+++ b/syncManager.py
@@ -1,0 +1,65 @@
+import threading
+from threading import Barrier, Event
+
+
+class SyncManager:
+    _instance = None
+    _lock = threading.RLock()
+
+    def __new__(cls, barrier_size):
+        with cls._lock:
+            if cls._instance is None:
+                cls._instance = super(SyncManager, cls).__new__(cls)
+                cls._initialize(barrier_size)
+        return cls._instance
+
+    @staticmethod
+    def _initialize(barrier_size):
+        SyncManager._instance.start_barrier = Barrier(barrier_size)
+        SyncManager._instance.client_finished = Event()
+        SyncManager._instance.server_alive = Event()
+
+    @classmethod
+    def reset(cls, barrier_size):
+        with cls._lock:
+            if barrier_size >= 0:
+                if cls._instance is None:
+                    cls.__new__(cls, barrier_size)
+                else:
+                    cls._initialize(barrier_size)
+
+    @classmethod
+    def client_finished(cls):
+        if cls._instance:
+            return cls._instance.client_finished
+
+    @classmethod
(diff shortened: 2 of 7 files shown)


validateOffload.py
```diff
@@ -118,9 +118,9 @@ def output(self, out: TftAggregateOutput):
 
         if self.iperf_pod_type == PodType.HOSTBACKED:
             if isinstance(self._iperf_instance, IperfClient):
-                logger.info(f"The client VF representor ovn-k8s-mp0_0 does not exist")
```
wizhaoredhat, line 122: It should still be there. Did you test this with ovs HWOL?

validateOffload.py
```diff
@@ -102,7 +104,8 @@ def stat(self, duration: int) -> Result:
             r1 = self.run_ethtool_cmd(self.ethtool_cmd)
             if r1.returncode != 0:
                 return r1
-            time.sleep(duration)
```
wizhaoredhat: Good catch during the presentation. I really like your fix here :+1:

measurePower.py
```diff
@@ -37,29 +38,24 @@ def extract(r: Result) -> int:
...
             iteration = 0
-            while True:
+            while SyncManager.client_not_finished():
                 r = self.run_oc(cmd)
                 if r.returncode != 0:
                     logger.error(f"Failed to get power {cmd}: {r}")
                 pwr = extract(r)
                 total_pwr += pwr
                 iteration += 1
-                # FIXME: Hardcode interval for now
```
wizhaoredhat, line 50: This is a bit nuanced the reason the FIXME is here is because we want to measure power strictly between the start and stop of the test. E.g.
```
       Start Tests                                     Stop Test
|---------|--------------------------------------------|-------|
     |<---- Measuring power here is not desirable
                                |<------ Measuring power here is desirable
  So what could happen is that you are measuring power here:
                                                          |<---- just a few ms after the client finishes but when you checked before it didn't finish
```
So it might be better to have another sync right before the client is about to end. I know this power measurement is a bit tricky because it measures instantaneous power instead of average power (I believe). If it is average power then it should be ok. E.g.
```
Instantaneous power reading: 162 Watts
Minimum during sampling period: 54 Watts
Maximum during sampling period: 324 Watts
Average power reading over sample period: 132 Watts
```

validateOffload.py
Zinkelburger: It no longer has the `_0` suffix

measurePower.py
Zinkelburger: I'm not sure how to sync before the client is about to end.

Maybe we can log the timestamp of when the client finishes, and discard measurements after that? We could do this for MeasureCPU as well

validateOffload.py
Zinkelburger: Thank you :wink:

measurePower.py
wizhaoredhat: Do you think a semaphore will work here? Or we using the inbuilt timeout functionality of iperf?

Zinkelburger: My timing idea is bad. Then we'd have to make sure to synchronize the times between all the pods, and it would be very fragile/sometimes fail for no reason.

I changed measurePower to measure every 0.2 seconds instead of 2 seconds. Because there are so many more measurements, we can probably afford to simply discard the last few power results if they affect the average too much.

I can take a look at how the last few measurements affect the average, if this sounds reasonable.

wizhaoredhat: I think this is ok for now.

syncManager.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @staticmethod
+    def _initialize(barrier_size):
+        SyncManager._instance.start_barrier = Barrier(barrier_size)
+        SyncManager._instance.client_finished = Event()
+        SyncManager._instance.server_alive = Event()
+
+    @classmethod
+    def reset(cls, barrier_size):
+        with cls._lock:
```
SalDaniele, line 24: "assert barrier_size > 0"

Add a clear error message if we ever hit this case

syncManager.py
```diff
@@ -0,0 +1,65 @@
...
+        if cls._instance:
+            cls._instance.client_finished.set()
+
+    @classmethod
+    def set_server_alive(cls):
+        if cls._instance:
+            cls._instance.server_alive.set()
+
+    @classmethod
+    def wait_on_barrier(cls):
```
SalDaniele, line 47: Can we add a comment here explaining the behavior and how threads will decrement the barrier count to ensure they start "together".

Code makes sense to me, but just to make sure we remember how this works in 12 months :)

syncManager.py
```diff
@@ -0,0 +1,65 @@
...
+    @classmethod
+    def reset(cls, barrier_size):
+        with cls._lock:
+            if barrier_size >= 0:
+                if cls._instance is None:
+                    cls.__new__(cls, barrier_size)
+                else:
+                    cls._initialize(barrier_size)
+
+    @classmethod
```
SalDaniele, line 31: We can remove this

SalDaniele:
LGTM

PR #23, opened 2024-04-25
Remove unused `-t`

Zinkelburger:
When viewing the logs of the pod, they contain things like "can't open /dev/tty2: No such file or directory"

- Remove the unneeded `-t` in `oc exec`
- I was able to run the tests with this change


wizhaoredhat:
LGTM

PR #24, opened 2024-04-26
Cumulative lint

Zinkelburger:
Cumulative lint, including changes from my unmerged PRs.

Cherry picked commits from my other lint branch. I believed I have removed most of the conflicting commits. 
I'm here Monday/Tuesday, so I'll continue to clean it up .

I still have some suggestions from Thomas to address, but I was able to run this successfully on 227.


common.py
```diff
@@ -114,14 +148,60 @@ class TftAggregateOutput:
...
+    elif isinstance(data, list):
+        return [serialize_enum(item) for item in data]
+    else:
+        return data
+
+
+T = TypeVar("T")
+
+
+def dataclass_from_dict(cls: Type[T], data: Dict[str, Any]) -> T:
```
SalDaniele, line 195: Lets add a comment explaining how this function works as well

mypy.ini
```diff
@@ -0,0 +1,2 @@
+[mypy]
```
SalDaniele, line 1: We can lint these too, lets not add this to this PR though

testConfig.py
```diff
@@ -14,21 +15,19 @@ class ClusterMode(Enum):
...
-            self.fullConfig = safe_load(io.StringIO(contents))
-
-        self.kubeconfig_tenant = "/root/kubeconfig.tenantcluster"
-        self.kubeconfig_infra = "/root/kubeconfig.infracluster"
-        self.kubeconfig_single = "/root/kubeconfig.nicmodecluster"
-        self.kubeconfig_cx = "/root/kubeconfig.smartniccluster"
-        self.client_tenant = None
-        self.client_infra = None
-        self.server_node = None
-        self.client_node = None
```
SalDaniele, line 31: Why are we deleting this fields? Are they not needed?

SalDaniele:
LGTM, a couple nits, just ping me when this is ready to merge

common.py
Zinkelburger: [Sure thing](https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/24/files#diff-306e28897a4282ae3852eb0568483edafeed04a979ec5b8afa2b2ee685f2148dR195)

testConfig.py
Zinkelburger: Client_node and server_node were not used. 

Tenant and infra are initialized as before, that part of the code [remains unchanged](https://github.com/wizhaoredhat/ocp-traffic-flow-tests/blob/main/testConfig.py#L36)

Only issue could be having them uninitialized instead of `None` . Although I could add some `asserts`, I believe this [guarantees that they will be initialized](https://github.com/wizhaoredhat/ocp-traffic-flow-tests/blob/main/testConfig.py#L52):
```python        
else:
        logger.error("Cannot Find Kubeconfig.")
        sys.exit(-1)
```

PR #28, opened 2024-05-10
Add hwol evaluation

SalDaniele:
update evaluator to determine the success/failure of the hwol validation
plugin.


SalDaniele:
@Zinkelburger @vrindle PTAL

evaluator.py
```diff
@@ -194,10 +272,21 @@ def evaluate_pass_fail_status(self) -> PassFailStatus:
                 total_passing += 1
```
Zinkelburger, line 272: you may want to rename these to `tft_passing`, `tft_failing` to be consistent in this block:
```python
num_tft_passed=total_passing,
num_tft_failed=total_failing,
num_plugin_passed=plugin_passing,
num_plugin_failed=plugin_failing
```

validateOffload.py
```diff
@@ -136,6 +143,9 @@ def output(self, out: TftAggregateOutput) -> None:
         )
 
     def generate_output(self, data: str) -> PluginOutput:
+        # Different behavior has been seen from the ethtool output depending on the driver in question
+        # Log the output of ethtool temporarily until this is more stable.
+        logger.info(f"generate hwol output from data: {data}")
```
Zinkelburger, line 148: I think I had this while debugging at one point, but it was very verbose. Maybe `logger.debug` is more appropriate

validateOffload.py
```diff
@@ -81,6 +81,12 @@ def run_ethtool_cmd(self, ethtool_cmd: str) -> Result:
         return r
 
     def parse_packets(self, output: str, packet_type: str) -> int:
+        prefix = f"{packet_type}_packets"
+        if prefix in output:
+            for line in output.splitlines():
+                stripped_line = line.strip()
+                if stripped_line.startswith(prefix):
+                    return int(stripped_line.split(":")[1])
```
Zinkelburger, line 89: Smart change, my bad removing that.
The function is definitely more complicated now, it could make sense to refactor it.
Or maybe just add a comment like "case 1: rx_packets is in the ethtool output" "case 2: manually summing the queues"

Zinkelburger (commented):
Small nits but LTGM

evaluator.py
SalDaniele: Done

validateOffload.py
SalDaniele: Done

validateOffload.py
SalDaniele: I agree, I added a comment to switch this to debug  once this part of the code is stable. I want to make sure we catch the behavior you described in the jenkins logs if ethtool does not provide the expected fields.

tests/test_evaluator.py
```diff
@@ -25,9 +25,7 @@ def run_subprocess(command, **kwargs):
 def test_evaluator_valid_input() -> None:
     log_path = os.path.join(current_dir, "input1.json")
     compare_path1 = os.path.join(current_dir, "output1a.json")
-    compare_path2 = os.path.join(current_dir, "output1b.txt")
```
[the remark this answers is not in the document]
Zinkelburger: That makes sense to me

.github/workflows/lint-test.yml
```diff
@@ -25,7 +25,7 @@ jobs:
     - name: Check code formatting with Black
       run: |
        black --version
-       black --check .
+       black --check --diff .
```
vrindle, line 28: OOC: What is the reasoning for changing this line?

evaluator.py
```diff
@@ -107,14 +161,26 @@ def eval_log(self, log_path: Path) -> None:
...
 
             self._eval_flow_test(run["flow_test"])
-            for plugin in run["plugins"]:
-                # TODO: add evaluation for plugins
-                # logger.debug(f'Reading result from plugin {plugin["name"]}')
-                pass
+            for plugin_output in run["plugins"]:
+                # TODO: add evaluation for measure_cpu and measure_power plugins
+                plugin_output = dataclass_from_dict(PluginOutput, plugin_output)
+                if plugin_output.name == "validate_offload":
```
vrindle, line 167: could we put string literal "validate_offload" in constant or enum?

evaluator.py
```diff
@@ -187,17 +265,28 @@ def calculate_gbps_http(self, result: dict) -> Bitrate:
...
+            if plugin_result.success:
+                plugin_passing += 1
+            else:
+                plugin_failing += 1
+
         return PassFailStatus(
-            result=total_failing == 0,
-            num_passed=total_passing,
-            num_failed=total_failing,
+            result=tft_failing + plugin_failing == 0,
```
vrindle, line 285:
```🜛 ~ python3
Python 3.12.3 (main, Apr  9 2024, 08:09:14) [Clang 15.0.0 (clang-1500.3.9.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> True + 1
2
```
In python booleans can be added to integers so let's make sure we doing the addition first (would make sense to add parenthesis in this case)

tests/test_evaluator.py
```diff
@@ -36,22 +34,11 @@ def test_evaluator_valid_input() -> None:
         check=False,
     )
 
-    with open(output_path2, "w") as printed_output_file:
```
vrindle, line 39: What's the reasoning for getting rid of this part?

.github/workflows/lint-test.yml
SalDaniele: https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#:~:text=echo%20%24%3F%0A123-,%2D%2Ddiff,-%C2%B6

When we push changes, it isnt helpful to just know that black failed, we want to see the diff that black would apply to fix the issue

tests/test_evaluator.py
SalDaniele: This is provided in the commit message. i.e.

> We should not attempt a 1-to-1 match of stdout from running the code.
> This is not an API that we need to expect consistent output from (i.e.
> someone may change a print statement, this should not break tests).
> 
> Only test on the json output of the program.

evaluator.py
SalDaniele: There is operator precedence in python so this will not happen:

https://www.programiz.com/python-programming/precedence-associativity#:~:text=The%20operator%20precedence%20in%20Python%20is%20listed%20in%20the%20following%20table.%20It%20is%20in%20descending%20order%20(upper%20group%20has%20higher%20precedence%20than%20the%20lower%20ones).

Also, these are not booleans, they are integers.

evaluator.py
SalDaniele: Done

tests/test_evaluator.py
SalDaniele: https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/28#discussion_r1597217193

PR #34, opened 2024-05-15
Add Default Network

wizhaoredhat:


SalDaniele:
Nice, LGTM. Maybe we can just add an update to the nice README you added to include this field when running sr-iov testing.

PR #35, opened 2024-05-16
Refactor Iperf into another inherited class of PerfServer/Client

wizhaoredhat:


validateOffload.py
```diff
@@ -11,7 +11,8 @@
 from logger import logger
 import time
 from testConfig import TestConfig
-from iperf import IperfServer, IperfClient, EXTERNAL_IPERF3_SERVER
+from iperf import IperfServer, IperfClient, EXTERNAL_PERF_SERVER
```
SalDaniele, line 14: EXTERNAL_PERF_SERVER is imported from perf now

iperf.py
```diff
@@ -68,27 +33,6 @@ def __init__(self, tc: TestConfig, ts: TestSettings):
         self.cluster_ip_addr = self.create_cluster_ip_service()
         self.nodeport_ip_addr = self.create_node_port_service(self.port + 25000)
```
SalDaniele, line 34: We need to set these in the perf class as well if there is code in the perf class that expects this fields to exist

SalDaniele (commented):
LGTM besides some MyPy errors.

More generally, curious why an additional layer of abstraction between Task and Iperf is required. I support the netperf task will share a significant amount of code with Iperf that isnt already included in the base Task class?

validateOffload.py
wizhaoredhat: Thanks for catching this.

iperf.py
wizhaoredhat: Yep I agree. Made the change.

PR #36, opened 2024-05-22
Add Netperf Testing

wizhaoredhat:
- Also removed the requirement to pass in the evaluator config

### DIFF
--- a/netperf.py
+++ b/netperf.py
@@ -1,0 +1,125 @@
+import common
+from common import PodType, ConnectionMode, TestType, IperfOutput
+from logger import logger
+from testConfig import TestConfig
+from thread import ReturnValueThread
+from task import Task
+from common import Result
+from testSettings import TestSettings
+import json
+import time
+from syncManager import SyncManager
+import sys
+import perf
+
+NETPERF_SERVER_EXE = "netserver"
+NETPERF_CLIENT_EXE = "netperf"
+
+
+class NetPerfServer(perf.PerfServer):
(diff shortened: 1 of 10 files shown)


netperf.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if self.reverse:
+            logger.info(f"Reverse is not supported by Netperf")
+
+        logger.info(f"Running {self.cmd}")
+
+        self.exec_thread = ReturnValueThread(target=client, args=(self, self.cmd))
+        self.exec_thread.start()
+
+    def generate_output(self, data: str) -> IperfOutput:
```
SalDaniele, line 83: IperfOutput -> PerfOutput

netperf.py
```diff
@@ -0,0 +1,125 @@
...
+
+    def output(self, out: common.TftAggregateOutput) -> None:
+        # Return machine-readable output to top level
+        assert isinstance(
+            self._output, IperfOutput
+        ), f"Expected variable to be of type IperfOutput, got {type(self._output)} instead."
+        out.flow_test = self._output
+
+        # Print summary to console logs
+        logger.info(f"Results of {self.ts.get_test_str()}:")
```
SalDaniele, line 125: Maybe it would be helpful for debugging to print a summary of the output here as we do for iperf tests.

SalDaniele (commented):
LGTM

netperf.py
wizhaoredhat: We will need to refactor this a bit to handle different result outputs from iperf and netperf. We can do this later.

PR #40, opened 2024-05-28
[th/more-typing] use a stricter `mypy` check and fix typing errors for that

thom311:
Currently, our github action runs `mypy --install-types --non-interactive --disallow-untyped-defs --ignore-missing-imports .`. 

We will adjust that to run `mypy --strict --config-file mypy.ini .` which is stricter. Fix the code for the typing errors due to that.

Also replace `typing.{List,Dict,Tuple,Union}`.

### DIFF
--- a/common.py
+++ b/common.py
@@ -1,7 +1,7 @@
 import jinja2
 from dataclasses import dataclass, fields, field, is_dataclass
 from enum import Enum
-from typing import Optional, Any, Union, Type, TypeVar, cast
+from typing import Optional, Any, Type, TypeVar, cast
 from typing import Mapping
 
 TFT_TOOLS_IMG = "quay.io/wizhao/tft-tools:latest"
@@ -22,7 +22,7 @@ class Result:
 E = TypeVar("E", bound=Enum)
 
 
-def enum_convert(enum_type: Type[E], value: Union[E, str, int]) -> E:
+def enum_convert(enum_type: Type[E], value: E | str | int) -> E:
     if isinstance(value, enum_type):
         return value
     elif isinstance(value, str):
@@ -132,7 +132,7 @@ class TestMetadata:
 @dataclass
 class BaseOutput:
     command: str
-    result: dict[str, Union[str, int]]
+    result: dict[str, str | int]
 
     def __init__(self, command: str, result: Mapping[str, str | int]):
         if not isinstance(result, dict):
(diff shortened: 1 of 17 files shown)


common.py
```diff
@@ -110,30 +110,53 @@ class TestMetadata:
...
     client: PodInfo
 
-    def __post_init__(self) -> None:
-        self.test_case_id = enum_convert(TestCaseType, self.test_case_id)
-        self.test_type = enum_convert(TestType, self.test_type)
-        if isinstance(self.server, dict):
-            self.server = dataclass_from_dict(PodInfo, self.server)
-        if isinstance(self.client, dict):
-            self.client = dataclass_from_dict(PodInfo, self.client)
+    def __init__(
```
wizhaoredhat, line 113: why was this changed from post_init to init?

thom311: in a simplified form, the code previously was
```python
@dataclass
class TestMetadata:
    test_type: TestType
    def __post_init__(self):
        self.test_type = enum_convert(TestType, self.test_type)
```

In this case, `TestMetadata` has signature `test_type: TestType`. It makes no sense to afterwards in `__post_init__()` expect that it also could be a `str | int` and do an `enum_convert()`.

That inconsistency can be resolved in two ways:

```python
class TestMetadata:
    test_type: TestType
```
here you don't need an `__post_init__()`. All callers must pass a `TestType` argument. If they have a string at hand, the callers must convert it.

Instead I did:
```python
class TestMetadata:
    test_type: TestType
    def __init__(self, test_type: TestType | str | int):
        self.test_type = enum_convert(TestType, test_type)
```
Now you can construct the instance with a `test_type: TestType | str | int`, but `instance.test_type` will then be normalized/converted to an `TestType` enum.

Both works. The former pushes the work of converting the type to the (potentially many) callers. It also pushes the knowledge of how to get from a `str|int` to a `TestType` to those callers. I think it's better if that complexity is at one place.

Granted, few callers actually use this feature/flexiblity. Version 1 would work too.

wizhaoredhat: Got it, thanks!

k8sClient.py
```diff
@@ -1,4 +1,4 @@
-import kubernetes
+import kubernetes  # type: ignore
```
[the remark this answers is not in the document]
wizhaoredhat: The typing information for kubernetes python? I don't think so

wizhaoredhat:
@SalDaniele  PTAL.
LGTM from myside.

SalDaniele:
LGTM

PR #41, opened 2024-06-03
[th/plugin-class] introduce a common baseclass "pluginbase.Plugin" and some refactoring

thom311:
The plugins should have a base class, so we can treat them in a generic way.

Add "pluginbase.Plugin" and plugin-specific subclasses.

The users no longer need to know the plugin. They can call `pluginbase.get_by_name(plugin_name)` to get a plugin, and then they can get everything they need from the abstract class, without knowing the plugin explicitly.

Also, move some code around. For example, move code specific to "ocp-traffic-flow-tests" from "common.py" to a new module "tftbase.py". The benefit is that "common.py" only contains general purpose utilities (and theoretically could also be copy-and-pasted to another python project). Even if that copy-and-pasting never happens, this is a better organization of the code, where modules have a clearer purpose and a clearer order of dependencies.


common.py
```diff
@@ -1,199 +1,120 @@
...
-    err: str
-    returncode: int
+from dataclasses import fields
+from dataclasses import is_dataclass
+from enum import Enum
+from typing import Any
+from typing import Optional
+from typing import Type
+from typing import TypeVar
+from typing import cast
```
SalDaniele, line 10: OOC, why specify these one per line rather than in a single import? Is this enforced by some linter?

validateOffload.py
```diff
@@ -10,14 +12,83 @@
...
+        perf_client: perf.PerfClient,
+        tenant: bool,
+    ) -> list[Task]:
+        # TODO allow this to run on each individual server + client pairs.
+        return [
+            ValidateOffload(tc, perf_server, tenant),
+            ValidateOffload(tc, perf_client, tenant),
+        ]
+
+    def eval_log(
```
SalDaniele, line 53: I like the use of the abstract class to add plugin.enable.

I am not so sure about moving the evaluation logic into the plugin. Originally, the goal was to have this code separated into two parts:

- the flow test tool which takes in a config and creates a raw json output
- the evaluator which takes in that json output, parses out the meaningful data, and determines success/failure.

It seems like these roles would be less differentiated by moving the evaluation logic into the plugin class. WDYT?

SalDaniele (commented):
LGTM. Definitely helps with the maintainability / clarity of the code, thanks for the PR!

common.py
thom311: I try to give the reasoning in the commit message. If that is unclear, then at least the commit message should be improved. But let me explain here:

it's not enforced by a linter. There is https://pypi.org/project/flake8-import-order/ , however that does not enforce single-line imports. `flake8-import-order` is also not followed, because it has additional style suggestions, that I found cumbersome.

The only reason why I did this, is because IMO this is more readable and nicer with future changes (a diff adds/removes lines for the import that changes, nothing else). I guess, that opinion is not unversally shared, if you see https://stackoverflow.com/questions/15011367/python-module-import-single-line-vs-multi-line .

In short, just my preference. But we can adopt another style, if that is preferred.

validateOffload.py
thom311: for now, there is only one plugin-specific implementation of `eval_log()`. It's hard to guess, whether a new plugin will have widely different API, that makes it hard to abstract. In any case, when that plugin gets added, then it's easy to modify the code, so that it can again be good. Maybe something like:

```
    plugin = pluginbase.get_by_name(plugin_output.name)
    if plugin.eval_log_is_special_kind():
          plugin_result = plugin.eval_log_special_kind(special_args...)
    else:
          plugin_result = plugin.eval_log(plugin_output, run["flow_test"].tft_metadata)
```

in any case, for the currently existing code, this abstract well.

---

> the evaluator which takes in that json output, parses out the meaningful data, and determines success/failure.

This didn't change, did it?

All that the branch does at this place is replace
```python
        if plugin_output.name == VALIDATE_OFFLOAD_PLUGIN:
                    self._eval_validate_offload(...
```
with
```python
                plugin = pluginbase.get_by_name(plugin_output.name)
                plugin_result = plugin.eval_log(
                     plugin_output, run["flow_test"].tft_metadata
                 )
```

It really just moves code around. And having code specific to "validate_offload" plugin in `evaluate.py` is not nice.

common.py
SalDaniele: I see, sorry to make you repeat yourself, I missed the commit message. Makes sense to me, I like the multi-line

validateOffload.py
SalDaniele: Yeah, make sense to me

PR #42, opened 2024-06-04
[th/black-reformat] style: adjust code so that different black versions format it the same

thom311:
The formatting that python black produces depends on the version of the tool. The authoritative version is the one from the github action, that is, "24.4.2" which we currently install via pip.

On current Fedora 40, we have 22.12.0.

It's cumbersome if different versions fight over the formatting. Even if the correct solution is to use the same version as used in github, we can avoid this problem by rewriting the code a bit, so that it formats the same with versions 22.12.0 and 24.4.2.


wizhaoredhat:
LGTM

PR #43, opened 2024-06-04
[th/config-class] add strongly types, immutable data classes for TestConfig

thom311:
Instead of passing around the dictionary we parsed from YAML, create strongly typed classes that hold the parsed configuration.

Those classes are immutable, so it's easy to reason how they are used (always read-only).

Also, perform strict checking of the configuration and do so early when parsing the configuration. Also normalize the configuration to well-defined types and values. For example, plugins in YAML can be
```yaml
   - plugins:
       - name: plugin1
 ```
or simply strings:
```yaml
  - plugins:
    - plugin1
 ```

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -32,7 +32,6 @@ tft:
         instances: (7)
         server:
           - name: "(8)"
-            persistent: "(9)"
```
wizhaoredhat, line 35: This is useful for debugging. We should have a way to persist the server.

thom311: this is now implemented.

Note that there is no auto-detection of the flag. It defaults to FALSE, but is now configurable via the config YAML.

Previously it was always FALSE, but there was a code comment that would indicate that the default should somehow be autodetected. If that's useful, it's a future feature.

wizhaoredhat:
LGTM, just needs to be tested.

thom311:
> LGTM, just needs to be tested.

Some changes were pushed.

- `git diff 0a50df073a09dde3ebb6f14a3046f42877f6fc9d 41e883ee0cd6a4705b372ec7f005585f6e9e7580`
- `git range-diff 0a50df073a09dde3ebb6f14a3046f42877f6fc9d...41e883ee0cd6a4705b372ec7f005585f6e9e7580`

iperf.py
```diff
@@ -26,14 +26,14 @@
...
+        super().__init__(tc, ts)
+
+        self.exec_persistent = ts.conf_server.persistent
 
-        self.exec_persistent = ts.server_is_persistent
         if self.exec_persistent:
             self.template_args["command"] = IPERF_EXE
             self.template_args["args"] = f'["-s", "-p", "{self.port}"]'
 
-        perf.PerfServer.__init__(self, tc, ts)
```
wizhaoredhat, line 35: This was intentional because we need to have template_args set before calling the parent class's init.

thom311: ah. Well, it doesn't work, because at this point, `self.template_args` does not exist, and you'd get an exception. This change avoids the crash.

Indeed, `perf.PerfServer.__init__()` will already render the file, and it needs those template arguments. 

On [`th/next`](https://github.com/thom311/ocp-traffic-flow-tests/commits/th/next/) branch I already reworke that further (see commit "task: extract initialize() method").

I'd prefer if we could merge this commit as-is (at least it doesn't crash right away -- it only fails to template the values).

This will be fixed later (after this MR, with patches from `th/next` branch.

@wizhaoredhat OK?

wizhaoredhat: This is ok with me. Thanks!

thom311:
I did some testing. It seems to work well for me (well, it also requires #44, but that is a separate issue)

task.py
```diff
@@ -43,11 +43,7 @@ def __init__(
         self.tc = tc
 
     def run_oc(self, cmd: str) -> host.Result:
-        if self.tenant:
-            r = self.tc.client_tenant.oc(cmd)
-        else:
-            r = self.tc.client_infra.oc(cmd)
-        return r
+        return self.tc.client(tenant=self.tenant).oc(cmd)
```
wizhaoredhat, line 47: Shouldn't this be "=="?

common.py
```diff
@@ -45,74 +46,99 @@ def enum_convert(
...
+                    try:
+                        return int(s)
+                    except Exception:
+                        pass
+                    return cast(int, enum_convert(enum_type, s).value)
 
-            def _range_endpoint(s: str) -> int:
                 try:
-                    return int(s)
+                    # Try to detect this as range. Both end points may either by
```
wizhaoredhat, line 156: Please update the docs with the new way of listing test cases.

pluginbase.py
```diff
@@ -68,3 +59,15 @@ def get_by_name(plugin_name: str) -> Plugin:
     if plugin is None:
         raise ValueError(f'Plugin "{plugin_name}" does not exist')
     return plugin
+
+
+if typing.TYPE_CHECKING:
```
wizhaoredhat, line 80: Is this only for making the linter happy? If so... I want to avoid this weirdness.

pluginbase.py
```diff
@@ -1,37 +1,28 @@
...
-import perf
-
-from task import Task
-from testConfig import TestConfig
 from tftbase import PluginOutput
 from tftbase import PluginResult
 from tftbase import TestMetadata
 
 
-class PluginTask(Task):
```
wizhaoredhat, line 14: Why do we need to move this to Task? And my understanding is that this causes issues.

pluginbase.py
```diff
@@ -1,37 +1,28 @@
...
-
 class Plugin(ABC):
     PLUGIN_NAME = ""
 
     @abstractmethod
     def enable(
         self,
         *,
-        tc: TestConfig,
+        tc: "TestConfig",
```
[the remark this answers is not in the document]
wizhaoredhat: This just seems a very roundabout way to handle dependency issues.... Could we discuss any cleaner alternatives?

tests/test_common.py
```diff
@@ -19,6 +22,99 @@
...
+    assert common.str_to_bool("no") is False
+    assert common.str_to_bool("false") is False
+
+    # truthy
+    assert common.str_to_bool("1") is True
+    assert common.str_to_bool("y") is True
+    assert common.str_to_bool("yes") is True
+    assert common.str_to_bool("true") is True
+
+    assert common.str_to_bool("bogus", None) is None
```
wizhaoredhat, line 45: I don't understand this test case:
Should it be assert common.str_to_bool("bogus") is None. Why do we need to pass additional arguments to str_to_bool?

testSettings.py
```diff
@@ -1,56 +1,49 @@
...
         self,
-        connection_name: str,
+        connection: testConfig.ConfConnection,
         test_case_id: TestCaseType,
-        node_server_name: str,
-        node_client_name: str,
-        server_pod_type: PodType,
-        client_pod_type: PodType,
-        server_default_network: str,
-        client_default_network: str,
```
wizhaoredhat, line 22: I am confused where is "default_network" set or more specifically "conf_client.default_network"

task.py
thom311: no, this is correct. `tenant:bool` is a keyword argument, so this is like calling `return self.tc.client(tenant=True)`.

Except, in this case, the argument is not the `True` literal, it's `self.tenant` (which is also a bool).

common.py
thom311: ok. Note that the old way still works, so if you follow the (old) docs, you are still good.

pluginbase.py
thom311: it is. If you don't know what it means, you can ignore it.

But what you get by it, are useful linter checks that tell you when you do something wrong. And you should not ignore that.

pluginbase.py
thom311: The issue is the order of dependencies of certain modules, which this commit fixes.

It could also be fixed while leaving  `PluginTask` inside `pluginbase` module. But the issue is there regardless, and indeed, moving it doesn't cause nor fix the issue.

This commit moves the `PluginTask`, because it seems to make sense to have trivial types in `tftbase`. The triviality of those types, makes it a non-issue to have them at the very basic module (`tftbase`) and having them accessible to most other modules. Also, while `PluginTask` was previously beside the `Plugin` (in `pluginbase`), it's now beside the `Task` (in `tftbase`). So it's still beside a thing, to which it is closely related.

tests/test_common.py
thom311: if you don't pass a second argument, then `str_to_bool()` will raise an exception on unrecognized strings. That's a choice, by default it's strict in what it accepts.

This tests sets the `on_error` argument to `None`, which enables that any invalid argument is returned as the on-error value.

There is also a test
```python
    with pytest.raises(ValueError):
        common.str_to_bool("bogus")
```
which covers the case you mention.

testSettings.py
thom311: `conf_client` is the `testConfig.ConfClient` type, which is an object to keep the data from the test YAML, like
```
tft:
  - name: test1
    connections:
      - name: connection1
        client:
          - name: client1 # <<< HERE
            default-network: default/default
            sriov: True
```

This is parsed in testClient.py, see
```python
        default_network = "default/default"
        v = vdict.pop("default-network", None)
        if v is not None:
            if not isinstance(v, str):
                raise ValueError(f'"{yamlpath}.name": expects a string but got {name}')
            default_network = v
```

thom311: basically, search for `default_network` or `"default-network"` in `testConfig.py`. It comes from there.

And since all of `TestConfig` is immutable, it *only* gets initialized while creating the `TestConfig` instance, which happens early in `main.py` when we parse the YAML.

common.py
thom311: README.md adjusted.

common.py
```diff
@@ -10,6 +10,69 @@
...
+#
+# See also, dataclasses.MISSING and dataclasses._MISSING_TYPE
+class _MISSING_TYPE:
+    pass
+
+
+MISSING = _MISSING_TYPE()
+
+
+def bool_to_str(val: bool, *, format: str = "yes") -> str:
```
SalDaniele, line 27: Why would this function return "yes" and "no"? Shouldn't we return "true" "false"? I assume I am missing something here...

common.py
```diff
@@ -10,6 +10,69 @@
...
+T1 = TypeVar("T1")
+T2 = TypeVar("T2")
+
+
+def str_to_bool(
+    val: None | str | bool,
+    on_error: T1 | _MISSING_TYPE = MISSING,
+    *,
+    on_default: T2 | _MISSING_TYPE = MISSING,
+) -> bool | T1 | T2:
```
SalDaniele, line 42: I don't think I understand what is being added by having Missing_Type.

In my mind str_to_boot() should return a bool, and raise an exception if for some reason it is called with a str that cannot be intuitively converted to a bool.

Can you elaborate why this is necessary?

thom311: > In my mind str_to_boot() should return a bool, and raise an exception if for some reason it is called with a str that cannot be intuitively converted to a bool.

That's what it does. `common.str_to_bool("bogus")` raises an exception.

However, if you do `common.str_to_bool("bogus", on_error=something)` (or simply `common.str_to_bool("bogus", something)`, then on error `something` is returned.

the `_MISSING_TYPE` is there to get the typing information right.

- If you call `common.str_to_bool(str)`, then mypy knows that this (only) returns a `bool`. That means, the caller doesn't need to further check either.

- If you call `common.str_to_bool(str, something)`, then mypy knows that this returns `bool | type(something)`, consequently, the caller will be forced to deal with that. For example,

For example,
```python
    var: bool
    var = common.str_to_bool(str, None)

    ternary_val: Optional[bool]
    ternary_val = common.str_to_bool(str, None)
```
both works, as you would would expect. And mypy also will agree with this.

For example
```python
   def func(val: bool)
       pass

   func(common.str_to_bool(str, None))
```
will be rejected by the linter, because you didn't handle the `None`.

common.py
thom311: That's what the format is there for. Currently only `format="yes"` is implemented, which gives a `"yes"`/`"no"`. That's also the default format.

The most important part is that the output of `bool_to_str()` can be parsed back by `str_to_bool()`.

For no strong reasons, this defaults (and only implements) `format="yes"`. It could also implement `format="true"` (to get `"true"/"false") or even default to that.

Indeed, what would be a nice additional property, is that the string output of `bool_to_str()` can be parsed as a boolean in YAML. Indeed, "yes"/"no" is not valid YAML. That would be a reason to change this. I will change that.

SalDaniele: Sounds good to me

SalDaniele:
LGTM. Really like the changes to add "_test_case_typ_infos", this is much better.

testSettings.py
wizhaoredhat: Thanks, I think those changes were made in another commit that was why I was confused.

tests/test_common.py
wizhaoredhat: Can we add that test instead? I think it is more useful instead of using on_error.

pluginbase.py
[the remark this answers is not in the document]
wizhaoredhat: If we had a ABC for perf and others, would it solve some of the circular dependency issues? Also if we move some of the classes to their own files?

WDYT?

pluginbase.py
wizhaoredhat: Let's talk in the other thread about removing the circular dependency.

pluginbase.py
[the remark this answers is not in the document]
thom311: The patch solves a loop of the "import". That was an actual problem.

Beyond that, we still have a loop, for the typing hints and which types are referred to.

Maybe you could do something with moving classes around, and introducing new ABC classes (that have fewer dependencies themselves). I doubt it however, because implementations of the virtual `pluginbase.Plugin.enable()` want an actual `perf.PerfServer` type and not `perf_server: PerfServerABC`.

And if you think, "but I don't understand why there is `if typing.TYPE_CHECKING: from task import PerfServer`", then just try to remove it. Then you will note that mypy will complain about an issue, and you will make it obvious why it's there. Maybe then you can come up with solutions, that move code around, introduce an ABC and involve fewer/no `if typing.TYPE_CHECKING:`. That might be interesting (if possible), but there is no real problem solved by that. The current solution is IMO fine, it has no problematic import loop and typing hints fully work.

See also:
- https://mypy.readthedocs.io/en/stable/runtime_troubles.html#import-cycles
- https://www.reddit.com/r/learnpython/comments/zzgkxj/comment/j2bmhlj/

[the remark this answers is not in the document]
thom311: ah, `pluginbase.PluginTask` had to move to `task.PluginTask`, because it uses a `Task` for `class PluginTask(Task)`. it's not merely for typing hints, it's in actual use.

And, as we need to break the import loop between `pluginbase`->`task`->`testConfig`->`pluginbase`, something needs to move.

Since it makes sense that `testConfig` uses `pluginbase.Plugin` and that `task` uses `testConfig.TestConfig`, the chosen solution is that `pluginbase` does not import `task`. Hence, `PluginTask` cannot be in `pluginbase` and had to move (as it requires `task.Task` -- not only for typing hints).

pluginbase.py
thom311: see https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/43#discussion_r1635914486

(lets follow up there).

tests/test_common.py
thom311: Which test? I think that `test_str_to_bool()` covers *a lot* of cases (some even cases redundantly, which I don't mind).

If you think a case is missing, I don't see it.

common.py
thom311: changed to return `"true"/"false"`.

tests/test_common.py
wizhaoredhat: Ah sorry, I had to look further down. Yes I see that you added that case as well. Thanks! This is ok for me.

pluginbase.py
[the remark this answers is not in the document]
wizhaoredhat: Let's merge your PR in for now. I understand what you are trying to say. Maybe there is more work to decouple the dependencies and objects. But it is messy right now, not due to your changes but in general how we structured the inheritance.

wizhaoredhat:
LGTM

pluginbase.py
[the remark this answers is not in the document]
thom311: @wizhaoredhat thank you.

> But it is messy right now, not due to your changes but in general how we structured the inheritance.

This particular place is IMO *not* messy and there is nothing to improve (other places are IMO messy). Of course, if you come up with a patch that proves me wrong, then great...

PR #44, opened 2024-06-11
[th/plugin-fixes] fix some result handling of the plugin

thom311:
the result handling of the tasks is IMO not good. There are no strong types, and it's quite ill defined what is returned. This should be improved later (by returning typed instances instead of strings or dictionaries or undefined content).

for now, two patches to workaround crashes.

### DIFF
--- a/pluginMeasureCpu.py
+++ b/pluginMeasureCpu.py
@@ -75,9 +75,8 @@ class TaskMeasureCPU(PluginTask):
 
     def output(self, out: TftAggregateOutput) -> None:
         # Return machine-readable output to top level
-        assert isinstance(
-            self._output, PluginOutput
-        ), f"Expected variable to be of type PluginOutput, got {type(self._output)} instead."
+        if not isinstance(self._output, PluginOutput):
+            return
         out.plugins.append(self._output)
 
         # Print summary to console logs
--- a/pluginMeasurePower.py
+++ b/pluginMeasurePower.py
@@ -84,7 +84,7 @@ class TaskMeasurePower(PluginTask):
                 iteration += 1
                 time.sleep(0.2)
 
-            r = Result(f"{total_pwr/iteration}", "", 0)
+            r = Result(json.dumps({"measure_power": f"{total_pwr/iteration}"}), "", 0)
             return r
 
         # 1 report at intervals defined by the duration in seconds.
@@ -101,7 +101,7 @@ class TaskMeasurePower(PluginTask):
         out.plugins.append(self._output)
 
         # Print summary to console logs
-        logger.info(f"measurePower results: {self._output.result}")
+        logger.info(f"measurePower results: {self._output.result['measure_power']}")
 
     def generate_output(self, data: str) -> PluginOutput:
         parsed_data = json.loads(data)
--- a/pluginMeasurePower.py
+++ b/pluginMeasurePower.py
@@ -95,9 +95,8 @@ class TaskMeasurePower(PluginTask):
 
     def output(self, out: TftAggregateOutput) -> None:
         # Return machine-readable output to top level
-        assert isinstance(
-            self._output, PluginOutput
-        ), f"Expected variable to be of type PluginOutput, got {type(self._output)} instead."
+        if not isinstance(self._output, PluginOutput):
+            return
         out.plugins.append(self._output)
 
         # Print summary to console logs
--- a/pluginValidateOffload.py
+++ b/pluginValidateOffload.py
@@ -200,9 +200,9 @@ class TaskValidateOffload(PluginTask):
         self.exec_thread.start()
 
     def output(self, out: TftAggregateOutput) -> None:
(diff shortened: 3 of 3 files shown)


wizhaoredhat:
LGTM

pluginMeasureCpu.py
```diff
@@ -75,9 +75,8 @@ def stat(self: TaskMeasureCPU, cmd: str) -> Result:
 
     def output(self, out: TftAggregateOutput) -> None:
         # Return machine-readable output to top level
-        assert isinstance(
-            self._output, PluginOutput
-        ), f"Expected variable to be of type PluginOutput, got {type(self._output)} instead."
+        if not isinstance(self._output, PluginOutput):
+            return
```
SalDaniele, line 79: Shouldn't we raise an exception in this case?

thom311: The assertion (failure) already raises a kind of exception, which breaks things. So raising another kind of exception solves nothing (unless you also suggest, that the caller catches such exceptions and define exceptions to be something that can regularly happen and that callers should expect/handle -- that sounds cumbersome though).

The assertion failure happens, because the earlier run action failed to generate a result. There is really nothing to handle at this point, so ignoring is kinda fine.

The better solution will be, to handle such error cases in a cleaner way. That will happen later .But for now, this works around the crash (due to the assert *exception*).

SalDaniele: I guess what I am suggesting is that a crash is a reasonable thing to happen at this point if a situation occurs where generate_output failed to generate PluginOutput.

How would the evaluator know the difference between a plugin failing in this manner or a plugin not being installed in the first place if the end result of both is out.plugins does not contain the output from the run?

SalDaniele: I do agree the way we handle plugin results needs work in general. It looks like the success or failure of the plugin is evaluated here by checking whether stop() managed to call generate_output() or fell back on using a default BaseOutput() which seems like a pretty bad way to determine something has failed. Maybe the real fix is to address the issue here instead.

```
        if self.exec_thread.result is not None:
            r = self.exec_thread.result
            if r.returncode != 0:
                logger.error(
                    f"Error occurred while stopping {class_name}: errcode: {r.returncode} err {r.err}"
                )
            logger.debug(f"{class_name}.stop(): {r.out}")
            self._output = self.generate_output(data=r.out)
        else:
            logger.error(f"Thread {class_name} did not return a result")
            self._output = tftbase.BaseOutput("", {})
```

thom311: The code you show already "handles" the error by logging an ERROR message. How else would you like to handle it, a hard crash towards the end of the test?

In any case, the crash currently happens later, for indirect reasons.

This patch is a workaround, to get a test "passing" at all (without a crash). I am currently heavily reworking this code, so this will be improved later.

SalDaniele: > How else would you like to handle it, a hard crash towards the end of the test?

In my opinion the best way to handle unexpected exceptions during the tests runs would be to write some artifact to the resulting json file that the evaluator can take in and interpret as a plugin failure.

> This patch is a workaround, to get a test "passing" at all (without a crash). I am currently heavily reworking this code, so this will be improved later.

Sounds good, I am fine with this going in for now.

wizhaoredhat: @thom311 How are you getting in a crash state. This was working for me before.

thom311: @wizhaoredhat [here](http://file.brq.redhat.com/~thaller/ocp-tft-pr44.log) is a logfile for a testoutput. It runs commit c1714876ab148f00e30cf9ecddb4dc0b99d3e2b7. The logfile also shows the used config.yaml.

wizhaoredhat: Ah I see it is probably because that particular plugin "validateoffload" is broken

wizhaoredhat: I think we can fix validate offload. The assertion did it's job of telling us there is a bug.

thom311: > I think we can fix validate offload. The assertion did it's job of telling us there is a bug.

A large rework is in progress, but this issue means, I cannot successfully test `main` branch and compare the run to WIP branches.

Also, the tool handles already such this failure at two distinct earlier places (right after the error happens in the plugin, and later, when the Task notices that there is no result returned). It does so by logging error messages. Granted, a better error handling shall follow, so that the failure is also recorded in the output, but at this point, an additional assertion crash is not helpful. It just breaks things, when the logging clearly indicates that there is a problem.

IMO this workaround should be merged.

wizhaoredhat: OK sounds good to me. Ok to merge.

wizhaoredhat:
LGTM

pluginMeasureCpu.py
thom311: thank you!

PR #45, opened 2024-06-11
[th/log-format] logger: log the thread id and align level

thom311:
Since we have multiple threads, its useful to see which thread logs what.

Also, align the logging level. Otherwise, lines with different level misalign and are harder to read.


wizhaoredhat:
LGTM

PR #47, opened 2024-06-12
[th/strict-dataclasses] stricter check attributes for dataclasses and make them frozen (immutable)

thom311:
- add `@strict_dataclass` decorator, which can be used to automatically check that a dataclass was initialized with values of the expected type.
- expand `dataclass_from_dict()` to perform the same type checking.
- at the same time, `dataclass_from_dict()` can also handle certain conversions, like string to enum (which is useful, when we load the dict from YAML). This allows us to drop certain `__init__()` methods that worked around this problem.
- mark various dataclasses as frozen. Immutable types a great!!

### DIFF
--- a/common.py
+++ b/common.py
@@ -239,18 +239,57 @@ T = TypeVar("T")
 # Takes a dataclass and the dict you want to convert from
 # If your dataclass has a dataclass member, it handles that recursively
 def dataclass_from_dict(cls: Type[T], data: dict[str, Any]) -> T:
-    assert is_dataclass(
-        cls
-    ), "dataclass_from_dict() should only be used with dataclasses."
-    field_values = {}
-    for f in fields(cls):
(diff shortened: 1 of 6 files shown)


common.py
```diff
@@ -248,3 +251,101 @@ def dataclass_from_dict(cls: Type[T], data: dict[str, Any]) -> T:
...
+        # the original check.
+        _post_check = getattr(type(instance), "_post_check", None)
+        if _post_check is not None:
+            _post_check(instance)
+
+
+TCallable = typing.TypeVar("TCallable", bound=typing.Callable[..., typing.Any])
+
+
+def strict_dataclass(cls: TCallable) -> TCallable:
```
SalDaniele, line 381: This is really cool! Great addition

SalDaniele:
LGTM

PR #48, opened 2024-06-13
[th/podman-rm-fix] trafficFlowTest: fix podman-rm command line

thom311:
`podman rm --time n` fails unless we also specify "--force". If we do that, we don't need to call a separate "stop" first.

Also, currently Host.run() interprets the cmd string not as shell script, so specifying two commands separated by semicolon is not going to work. That probably should change, because interpreting the cmd as shell script is a useful feature. Anyway, for now, this doesn't work.

### DIFF
--- a/trafficFlowTests.py
+++ b/trafficFlowTests.py
@@ -88,7 +88,7 @@ class TrafficFlowTests:
         logger.info(
             f"Cleaning external containers {perf.EXTERNAL_PERF_SERVER} (if present)"
         )
-        cmd = f"podman stop --time 10 {perf.EXTERNAL_PERF_SERVER}; podman rm --time 10 {perf.EXTERNAL_PERF_SERVER}"
+        cmd = f"podman rm --force --time 10 {perf.EXTERNAL_PERF_SERVER}"
         self.lh.run(cmd)
 
     def _run_test_tasks(


trafficFlowTests.py
```diff
@@ -88,7 +88,7 @@ def _cleanup_previous_testspace(self, namespace: str) -> None:
         logger.info(
             f"Cleaning external containers {perf.EXTERNAL_PERF_SERVER} (if present)"
         )
-        cmd = f"podman stop --time 10 {perf.EXTERNAL_PERF_SERVER}; podman rm --time 10 {perf.EXTERNAL_PERF_SERVER}"
+        cmd = f"podman rm --force --time 10 {perf.EXTERNAL_PERF_SERVER}"
```
SalDaniele, line 91: Don't we lose graceful shutdown with this change? There is the potential that some resources may not be cleaned up properly.

Since we run the external server with "--rm" I believe we don't need the second part of the original command anyways, "podman stop --time 10 {perf.EXTERNAL_PERF_SERVER}" should forcible stop the container after 10 second grace period, and rm the image automatically.

thom311: In general, a process should almost never be stopped via SIGKILL, because that gives no opportunity for graceful shut down. Any sensible "stop" implementation, sends a SIGTERM first. The "--time" parameter would be there as fallback to send SIGKILL, if the process doesn't exit after being asked to stop.

I am not entirely sure that podman-rm does that. But doing anything else here seems quite broken and unexpected. Also, `man podman-rm` makes it sound like that (`--time` is "Seconds to wait before forcibly stopping the container."). How else could the `--time` option possibly work?

wizhaoredhat: Let's give this a try. If things break we can revert. It's a good thing that this only runs for a single test.

wizhaoredhat:
LGTM

PR #51, opened 2024-06-13
[th/no-exception-from-validateoffload] pluginValidateOffload: don't raise an exception when command fails

thom311:
An exception is not the right response. Don't raise an exception, because the caller also doesn't know what to do. We need to handle errors somehow where it happens.

Currently, we have a catch-all in ReturnValueThread.run(), which prevents the exception from crashing the thread. This is also wrong, because we must handle exceptions close to where we understand why they happen. Any unexpected exception is a bug, that don't need handling.

We already log an error message (which is some kind of handling).  We also return some Result, that shall indicate the failure.  Granted, this returning an error status via the result needs improvement (which will follow).

We certainly also need to address why this failure happens.

The first step, is to not raise wrong exceptions:
```
  2024-06-13 13:07:53 INFO    [th:140188990821952]: The VF representor is: e89042734463f0f
  2024-06-13 13:07:53 INFO    [th:140188990821952]: Running exec -n default tools-pod-worker-247-validate-offload -- /bin/sh -c "ethtool -S e89042734463f0f"
  2024-06-13 13:07:53 ERROR   [th:140188990821952]: Run_ethtool_cmd: Cannot get stats strings information: No such device
  command terminated with exit code 96
  , 96
  2024-06-13 13:07:53 ERROR   [th:140188990821952]: Thread with target <function TaskValidateOffload.run.<locals>.stat at 0x7f8062f0dbc0> experienced exception ValidateOffload error: Cannot get stats strings information: No such device
  command terminated with exit code 96
   returncode: 96
```
---

This is a followup to https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/44#discussion_r1635270038 , but it's still not the solution for why the failure happens.


SalDaniele:
LGTM

wizhaoredhat:
LGTM, @thom311 Please look into "exception from crashing the thread". In general we want errors to crash the test instead of letting it continue since it is not useful to continue with errors.

PR #52, opened 2024-06-14
[th/split-common-test] tests: split "tests/test_tftbase.py" out of "tests/test_common.py"

thom311:
"common.py" contains a bunch of general purpose helpers. Theoretically, those could be copied to other projects, and reused (reuse by copy and paste is often not great, but sometimes it's better than the alternatives).

As such, we should also be able to copy-and-paste the test. For that, "test_common.py" should be without extra tftbase dependency. Split the test.


wizhaoredhat:
LGTM

PR #53, opened 2024-06-14
[th/host-log-run] host: log for Host.run() and extend host.Host implementation

thom311:
a major rework of `Host.run()`. This is how I think it should be done.

- Most importantly, (automatically) log useful messages about the command invocation. Also have various (optional) arguments, to control the logging.

- Handle binary output in a sensible way.

For full details, see the commit message.

If you currently run debug logging, it doesn't really show much. With this patch, we can see all the commands that are invoked (and their output). Quote often, the commands that we invoke are rather interesting for understanding what happens, so this provides IMO a lot of useful debugging output for free.

### DIFF
--- a/host.py
+++ b/host.py
@@ -1,23 +1,84 @@
 import json
+import logging
 import os
 import shlex
 import subprocess
 import sys
+import threading
 import typing
 
 from abc import ABC
 from abc import abstractmethod
+from collections.abc import Iterable
+from collections.abc import Mapping
 from dataclasses import dataclass
 from typing import Any
+from typing import Optional
 
 from logger import logger
 
 
+_lock = threading.Lock()
+
+_unique_log_id_value = 0
+
+
+def _unique_log_id() -> int:
+    # For each run() call, we log a message when starting the command and when
+    # completing it. Add a unique number to those logging statements, so that
+    # we can easier find them in a large log.
+    with _lock:
+        global _unique_log_id_value
+        _unique_log_id_value += 1
+        return _unique_log_id_value
+
+
+T = typing.TypeVar("T", bound=str | bytes)
+
+
 @dataclass(frozen=True)
-class Result:
-    out: str
-    err: str
+class BaseResult(ABC, typing.Generic[T]):
+    out: T
+    err: T
     returncode: int
+
+    @property
+    def success(self) -> bool:
+        return self.returncode == 0
+
+    def debug_str(self) -> str:
+        if self.returncode == 0:
+            status = "succcess"
+        else:
+            status = f"failed ({self.returncode})"
+
+        out = ""
+        if self.out:
+            out = f"; out={repr(self.out)}"
+
+        err = ""
+        if self.err:
+            err = f"; err={repr(self.err)}"
+
+        return f"{status}{out}{err}"
+
+    def debug_msg(self) -> str:
+        return f"cmd {self.debug_str()}"
+
+
+@dataclass(frozen=True)
+class Result(BaseResult[str]):
+    pass
+
+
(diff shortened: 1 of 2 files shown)


host.py
```diff
@@ -1,24 +1,85 @@
...
 from logger import logger
 
 
+_lock = threading.Lock()
+
+_global_counter_val = 0
+
+
+def _global_counter() -> int:
+    # For each run() call, we log a message when staring the command and when
```
wizhaoredhat, line 27: Typo for "staring" -> "starting"
Maybe a more descriptive name than _global_counter would be better such as "_unique_log_id"? WDYT?

host.py
```diff
@@ -1,24 +1,85 @@
...
+        return f"cmd {self.debug_str()}"
+
+
+@dataclass(frozen=True)
+class Result(BaseResult[str]):
+    pass
+
+
+@dataclass(frozen=True)
+class BinResult(BaseResult[bytes]):
```
wizhaoredhat, line 75: What is an example of a binary result?

host.py
```diff
@@ -34,29 +95,217 @@ def all_ports(self) -> list[dict[str, Any]]:
         return typing.cast(list[dict[str, Any]], r)
 
     @abstractmethod
-    def run(self, cmd: str, env: dict[str, str] = os.environ.copy()) -> Result:
+    def pretty_str(self) -> str:
         pass
 
+    @typing.overload
+    def run(
```
wizhaoredhat, line 102: Why is there 2 "run" functions with the same arguments but different return value types?

wizhaoredhat: Also there is no code for these functions

host.py
thom311: there is one in the commit message:
```
python -c 'import host; r = host.local.run("printf \"foo:\\705\""); print(r)'
```
or really anytime, you get binary result, like 
```
$ python -c 'import host; host.local.run("echo -n hello | gzip")'
2024-06-18 19:47:05 DEBUG   [th:140307334326080]: cmd[1;localhost]: call `echo -n hello | gzip`
2024-06-18 19:47:05 ERROR   [th:140307334326080]: cmd[1;localhost]: └──> `echo -n hello | gzip`: [BINARY] [UNEXPECTED_BINARY] succcess; out=b'\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\x03\xcbH\xcd\xc9\xc9\x07\x00\x86\xa6\x106\x05\x00\x00\x00'
```

Since we call other processes (that we don't trust), we should not assume that we get benign UTF-8.

also, CDA has `Host.read_file()` which is implemented as `f"cat {filename}"`. That cannot read binary files.

Usually, we don't call programs that output binary data. The main point here is the logging, to catch cases where we get unexpected binary output.

host.py
thom311: it's for the `@typing.overload`.

https://docs.python.org/3/library/typing.html#overload

This describes certain combinations of signatures. In particular, this one exists so that mypy knows that `Host.run(..., text=False)` gives a binary `BinResult` and that `Host.run(..., text=True)` gives a string `Result`.

wizhaoredhat: It makes the code really hard to follow since it spans multiple lines and takes up a lot space in my opinion. This aspect of python typing makes the code bloated which I think python wanted to solve... ironically now it is even more bloated than C++.

host.py
wizhaoredhat: Ah i see. It is to catch unexpected cases.

wizhaoredhat:
LGTM.... But as I mentioned python typing is awful.

PR #54, opened 2024-06-14
[th/plugin-log-enable] plugin: log about enabling plugin

thom311:
It's useful to see in debug logs, which plugins get enabled (and when, and which Tasks they create).


wizhaoredhat:
LGTM

PR #55, opened 2024-06-14
[th/task-context] pass the TestSettings to the Task

thom311:
- instead of giving only the `TestConfig` to the `Task`, give it the `TestSettings`. This one contains the `TestConfig`, but more additional context about the current part of the test. As `TestSettings` is immutable, this is not a concern.

- add `ConfigDescriptor`, which is an immutable object that refers the (immutable) `TestConfig` and indexes about which part of the test we currently care about. For example, as we iterate over the `test_config.config.tft` list, the `ConfigDescriptor` tells is which the current index to the tft list is (and we can get correct `ConfTest` instance from it).

- with this, the task can know the currently configured namespace. The namespace is configurable in the config YAML, and now accessible at `cfg_descr.get_tft().namespace`. Previously, this setting was ignored.

- track the `evaluator_config` also in the `TestConfig` object. This is part of the run configuration, also add it there.

### DIFF
--- a/iperf.py
+++ b/iperf.py
@@ -9,7 +9,6 @@ import tftbase
 from host import Result
 from logger import logger
 from syncManager import SyncManager
-from testConfig import TestConfig
 from testSettings import TestSettings
 from tftbase import ConnectionMode
 from tftbase import IperfOutput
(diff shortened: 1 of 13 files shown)


wizhaoredhat:
LGTM

PR #56, opened 2024-06-14
[th/github-action-no-push] github: don't push test image for pull requests

thom311:
Pull requests don't have access to the quay secrets, this thus fails for certain users (or if the pull request originates from another repository, that is not quite clear to me).

Don't push the result, just build it.

---

Note how various the open pull requests have this task failed.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -6,7 +6,7 @@ on:
       - main
 
 jobs:
-  build-and-push:
+  build:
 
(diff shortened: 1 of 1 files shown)


SalDaniele:
@wizhaoredhat is it possible for us to add credentials for the quay repo to our github? Otherwise we can just stop build on PR, and just build on merges

wizhaoredhat:
I think this is ok. LGTM

PR #57, opened 2024-06-17
[th/perf-server-fix-initalize-external-ip] perf: fix PerfServer.initialize() with ConnectionMode.EXTERNAL_IP

thom311:
With "if connection_mode == ConnectionMode.EXTERNAL_IP:" we don't have a template file to render or apply. Fix PerfServer.initialize() for that case.

Fixes: 889791694752 ('task: cleanup initializing tasks')

### DIFF
--- a/perf.py
+++ b/perf.py
@@ -67,10 +67,14 @@ class PerfServer(Task):
 
     def initialize(self) -> None:
         super().initialize()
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
LGTM

PR #58, opened 2024-06-17
[th/podman-run-cleanup] perf: cleanup podman-run commands

thom311:
Adjust the two podman-run commands of iperf/netperf:

- don't overwrite "--entrypoint". That is commonly not necessary. Also note that we build our image "{tftbase.TFT_TOOLS_IMG} with the container file from "images/Containerfile". This doesn't specify ENTRYPOINT, so our image's entrypoint defaults to "/bin/sh -c", which is fine. The benefit is also that we don't need to split up the command and can re-use the same string for podman-run and kubectl-exec commands.

- pass "--init" to podman-run. It's kinda nice to have a mostly functioning init process inside our container. For example, processes with PID 1 won't receive SIGTERM if the don't handle it.

- pass "--replace". Usually, we shouldn't have the container with the requested name around. Still, it does not seem to hurt to be sure that we can start this container and replace the old one.

- extract the shared command line and use it for both podman-run and kubectl-exec.

- also, pass "--json" to podman-run when running the iperf command. That makes it do the same thing as when run via kubectl-exec.

### DIFF
--- a/iperf.py
+++ b/iperf.py
@@ -42,13 +42,14 @@ class IperfServer(perf.PerfServer):
         }
 
     def setup(self) -> None:
+        cmd = f"{IPERF_EXE} -s -p {self.port} --one-off --json"
         if self.connection_mode == ConnectionMode.EXTERNAL_IP:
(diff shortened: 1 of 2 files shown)


wizhaoredhat:
LGTM

PR #59, opened 2024-06-18
[th/misc] minor patches (use host.local, add unit test, use `collections.abc.Mapping`)

thom311:
3 unrelated patches, posted here together to have less overhead with the pull request.

- tests: add test for loading all Plugin
- all: use collections.abc.Mapping instead of deprecated typing.Mapping
- all: use host.local instead of creating new host.LocalHost() instances

### DIFF
--- a/evaluator.py
+++ b/evaluator.py
@@ -4,11 +4,11 @@ import sys
 import typing
 import yaml
 
+from collections.abc import Mapping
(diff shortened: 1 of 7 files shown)


wizhaoredhat:
Nice Thanks for the cleanup.
LGTM

PR #60, opened 2024-06-18
[th/host-no-ipa] host: drop host.Host.{ipa,ipr,all_ports}()

thom311:
They are currently unused, but they are also not great.

- it should not be a method of Host. Sure, it uses Host.run() to invoke a command, but that doesn't mean that all methods that invoke a command should be methods of the Host. Instead, let Host be about invoking a command, and if you want to add helper functions on top of that, then let them take a Host argument.

- the methods just pass untrusted output to json.loads(). We should have boundaries of trust, and calling to iproute2 should not be trusted to always give valid JSON. We should handle failure or non-JSON output. Granted, the "handling" is failing with some undefined exception. But most callers wouldn't care about this, and they shouldn't be bothered with having to add a try/except around each call.

- likewise, passing untrusted data to json.loads() just gives some dictionary with untrusted content. We would want that the output is validated and parsed and a well-known, typed object gets returned.

### DIFF
--- a/host.py
+++ b/host.py
@@ -1,4 +1,3 @@
-import json
 import logging
 import os
 import shlex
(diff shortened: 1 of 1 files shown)


thom311:
there is also a patch that would re-add such functions (`common.ip_links()`, `common.ip_addrs()`, `common.ip_routes()`).

This patch is not included here, because the code is currently unused. Not sure we want to include general-purpose helpers, unless they are really used. I personally would, but that's just here.

The patch is at https://github.com/thom311/ocp-traffic-flow-tests/commits/th/host-no-ipa-2/

wizhaoredhat:
LGTM

thom311:
The patch for adding API `common.get_{links,addrs,routes}()` is [here](https://gist.github.com/thom311/673638a341eccea9619890460e09197d).

PR #62, opened 2024-06-19
[th/traffic-flow-tests-no-state] remove state from TrafficFlowTests object and instead pass data as function arguments/return values

thom311:
Often, it is most convenient if we can attach some data to an object.

It can also make code harder to read, because this state is accessible from various places, and it's not clear in which order it gets modified. It can be better (if possible), to pass data around as function parameter. Then at any point, you know that the input-data comes from the caller, and you can look where the caller gets it from.

Drop all state from `TrafficFlowTests` object.

### DIFF
--- a/main.py
+++ b/main.py
@@ -64,8 +64,6 @@ def main() -> None:
 
     for cfg_descr in ConfigDescriptor(tc).describe_all_tft():
         tft.test_run(cfg_descr)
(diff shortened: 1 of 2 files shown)


SalDaniele:
LGTM, makes sense to remove the state from TrafficFlowTests objects since we can use the ConfigDescriptor class now to handle much of this.

PR #63, opened 2024-06-19
[th/test-settings-cleanup] testSettings: make TestSettings a frozen dataclass

thom311:
TestSettings is already immutable, in the sense that no code currently modifies a TestSettings instance after creation.

Make it a frozen dataclass, which helps enforcing the immutablility via typing hints and runtime checks.

Also, make the computed attributes to be property getters.  The benefit is, that it is quite obvious that these getters don't modify the object. More importantly, with the frozen data classes it would be cumbersome to set these values as attributes.

### DIFF
--- a/testSettings.py
+++ b/testSettings.py
@@ -1,3 +1,6 @@
+import dataclasses
+
+import common
 import testConfig
 import tftbase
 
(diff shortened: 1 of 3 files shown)


SalDaniele:
LGTM

PR #64, opened 2024-06-19
[th/struct-parse-helpers] testConfig: move helpers for parsing yaml to "common.structparse_*"

thom311:
These helpers are not specific to TestConfig. When we parse YAML (or structured data that comes from YAML or JSON), then may be useful to construct similar immutable data classes. Move the code from "testConfig.py" to "common.py" for reuse.


SalDaniele:
LGTM

PR #65, opened 2024-06-19
[th/dataclass-from-dict-extend] common: add dataclass_to_dict() and extend dataclass_from_dict()

thom311:
"dataclass_from_dict()" cannot read any kinds of data classes (and fields). It should however be able to read the classes we care about.

Extend it, so that Optional[dataclass] and list[dataclass] also works.

Also, add dataclass_to_dict(), which is the opposite operation.


wizhaoredhat:
LGTM

wizhaoredhat:
@SalDaniele PTAL

SalDaniele:
LGTM

PR #66, opened 2024-06-21
[th/tests-prettify-json] tests: prettify the JSON test inputs

thom311:
The format will change, and so will need to adjust those files. As they currently are one huge line, the diff becomes not reviewable.

Instead, prettify the files with `jq`.

This is no longer the original style that was written by the test. But as we don't need to test the JSON parser itself, this is equivalent.


wizhaoredhat:
LGTM

PR #67, opened 2024-06-21
[th/bitrate] tftbase: move Bitrate class

thom311:
This will be used later, when already the test run generates the Bitrate result. It will not only be used and known to the Evaluator.

Also, introduce NA (-1) for not available results. For example, TestType.HTTP won't measure the bit rate and it makes no sense.

Also, make it a frozen class and perform strict validation. For convenience, you can pass `int` as tx/rx value, but the constructor will convert them to floats.

### DIFF
--- a/evaluator.py
+++ b/evaluator.py
@@ -17,18 +17,12 @@ from common import dataclass_from_dict
 from common import serialize_enum
 from common import strict_dataclass
 from logger import logger
+from tftbase import Bitrate
 from tftbase import IperfOutput
 from tftbase import PluginOutput
 from tftbase import TFT_TESTS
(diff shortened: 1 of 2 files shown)


wizhaoredhat:
LGTM

wizhaoredhat:
PTAL @SalDaniele

thom311:
I will rework this patch. Expressing `NA` as `-1.0` is a kludge. It's better to make this an `Optional[float]` and unspecified rates are expressed with `None`. The benefit is that you get an error (from mypy and at runtime) if you try to compare None with a number.

Back on DRAFT.

SalDaniele:
LGTM

PR #69, opened 2024-07-01
[th/no-sync-manager] move locking primitives out of SyncManager to TestSettings

thom311:
SyncManager does some questionable things, trying to be a singleton object. Regardless of that, it's also unnecessary.

Those locking primitives are per-test. We have TestSettings which contain the per-test settings. It's the right place for the locking.

We also don't need to abstract the various accesses (`SyncManager.wait_on_client_finish()`). Those do very little, just expose `ts.event_client_finished` event directly. It has a handful of accesses (that you can easily grep for), and abstracting them is only another layer of code for little benefit.


SalDaniele:
LGTM

PR #71, opened 2024-07-02
[th/dataclasses-kw-only] tftbase: make result/output dataclasses kw_only

thom311:
The dataclasses are immutable, and thus the constructors have a larger number of arguments. For readability, callers should use keyword arguments. Set kw_only=True to force that.


wizhaoredhat:
LGTM

PR #72, opened 2024-07-03
[th/test-type-handler] handle TestType generically by adding a TestTypeHandler object

thom311:
We currently have 5 `TestType`s:

```python
class TestType(Enum):
    IPERF_TCP = 1
    IPERF_UDP = 2
    HTTP = 3
    NETPERF_TCP_STREAM = 4
    NETPERF_TCP_RR = 5
```

HTTP is not yet implemented (it will be soon), and `IPERF_{TCP,UDP}` is implmeneted by `iperf.py` and `NETPERF_TCP_{STREAM,RR}` is implemented by `netperf.py`.

Previously, the code would explicitly check at various places for the TestType enum, and then do one thing or another.

Instead, add a `testType.TestTypeHandler` object. You can get the handler with `testType.TestTypeHandler.get(test_type)`. This is an abstract base class, that will allow the callers to do whatever the need, without "knowing" the exact `TestType`. 

Also, rename files `iperf.py` and `netperf.py` to `testTypeIPerf.py` and `testTypeNetPerf.py`, respectively, to make it clear that those are implementations of the test-type.

Later we will add an implementation for `TestType.HTTP`. With this, the new type can be added by only introducing a `testTypeHttp.py` module, with minimal changes to the existing code.


wizhaoredhat:
LGTM

PR #73, opened 2024-07-03
[th/dataclass-to-json] common: add common.dataclass_to_json() helper

thom311:
This is useful for a quick way to serialize a class to string, for example to print a logging message.

Logging an object in JSON  can be more useful than str(obj), because the result can be deserialized (first with json.loads() and then common.dataclass_from_dict()).


wizhaoredhat:
LGTM. This will be used later?

PR #74, opened 2024-07-03
[th/task-rework] rework Task

thom311:
rework the Task implementation.

The main points are:

- `Task.generate_output()` is now merged into the async operation (`Task.start()` + `Task.stop()`). It also means, that the async operation no longer returns the intermediate `host.Result` (which previously had to be re-parsed by `Task.generate_output()`). The async operation now already returns the `BaseOutput` that we want.

- `Task.output()` is renamed to `Task.aggregate_output()`, which is more fitting name.

- `thread.ReturnValueThread` is renamed to `task.TaskOperation`. Which seems a better name. It's also now in the `task` module, because it returns a `BaseOutput` (which makes it no longer a thing that makes much sense outside of `task.Task`).


wizhaoredhat:
LGTM

PR #75, opened 2024-07-03
[th/eval-config] add EvalConfig and make Evaluator stateless

thom311:
- don't let the Evaluator object keep state. Instead, the few functions that are there take function arguments and return the result. The caller has to pass them around. Stateless objects are nice, because you don't need to think about their state. In particular, you see where an input argument comes from (from the caller) and you see what we do with the result (return it).

- add `evalConfig.Config` for parsing the `eval-config.yaml`. The benefit is tight validation of the configuration file and a strongly typed object that we can use. Currently the config  only contains the threshold. This will be more useful, if the configuration gets more complex (https://issues.redhat.com/browse/NHE-774)

- use `evalConfig.Config`.

### DIFF
--- a/evalConfig.py
+++ b/evalConfig.py
@@ -1,0 +1,166 @@
+from collections.abc import Mapping
+from dataclasses import dataclass
+from typing import Any
+
+import common
+import testType
+
+from common import StructParseBase
+from common import strict_dataclass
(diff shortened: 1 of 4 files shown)


wizhaoredhat:
LGTM

PR #76, opened 2024-07-05
[th/configure-test-image] allow selecting another test container image for testing

thom311:
The tests pull `tftbase.TFT_TOOLS_IMG`, which is set to `quay.io/wizhao/tft-tools:latest`. This image gets automatically rebuild on every push to `main` branch.

However, if you are testing a WIP branch that requires an update test image, it is cumbersome.

Allow overriding the test image by setting `TFT_TEST_IMAGE` environment variable. Also, allow overwriting the ImagePullPolicy by setting `TFT_IMAGE_PULL_POLICY` environment variable.

An alternative would be to make those settings part of the TestConfig. However, this option is only useful for testing/developing the flow tests, as such it seems more cumbersome to do it via the TestConfig. Also, you might think if those settings were part of the config YAML, then it would be well known which test image was used by looking at the config. However, we tend to use `:latest`, so a week after you rant the test, you also don't know anymore which test image was used. If you want to record which image was used when running the test a week ago, preserve the log output.

### DIFF
--- a/manifests/host-pod.yaml.j2
+++ b/manifests/host-pod.yaml.j2
@@ -17,7 +17,7 @@ spec:
     command:
       - "{{ command }}"
     args: {{ args }}
-    imagePullPolicy: IfNotPresent
+    imagePullPolicy: {{ image_pull_policy }}
     securityContext:
       capabilities:
         add: ["NET_ADMIN","NET_RAW"]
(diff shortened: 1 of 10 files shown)


wizhaoredhat:
LGTM
@thom311 Please add documentation on the readme on how to enable this.

PR #77, opened 2024-07-05
[th/http-test] http: implement test_type=HTTP for fetching a small file via HTTP

thom311:
Inspired by ovn-kuber-traffic-flow-tests (client: [1], server: [2]).

[1] https://github.com/Billy99/ovn-kuber-traffic-flow-tests/blob/17369edc2875b18c1c9df96683518cbc76600fe8/process-cmds.sh#L38
[2] https://github.com/Billy99/ovn-kuber-traffic-flow-tests/blob/17369edc2875b18c1c9df96683518cbc76600fe8/manifests/http-server-pod-v4.yaml.j2#L36

https://issues.redhat.com/browse/NHE-711

### DIFF
--- a/testTypeHttp.py
+++ b/testTypeHttp.py
@@ -1,0 +1,98 @@
+import json
+import shlex
+
+from dataclasses import dataclass
+
+import common
+import perf
+import tftbase
+
+from logger import logger
+from perf import PerfClient
+from perf import PerfServer
+from task import TaskOperation
+from testSettings import TestSettings
+from testType import TestTypeHandler
+from tftbase import BaseOutput
+from tftbase import IperfOutput
+from tftbase import TestType
+
+
+@dataclass(frozen=True)
+class TestTypeHandlerHttp(TestTypeHandler):
+    def __init__(self) -> None:
+        super().__init__(TestType.HTTP)
+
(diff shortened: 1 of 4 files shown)


testTypeHttp.py
```diff
@@ -0,0 +1,98 @@
...
+    def _create_task_operation(self) -> TaskOperation:
+        server_ip = self.get_target_ip()
+        cmd = f"exec {self.pod_name} -- curl --fail -s http://{server_ip}:{self.port}/data"
+
+        def _thread_action() -> BaseOutput:
+            self.ts.clmo_barrier.wait()
+            r = self.run_oc(cmd)
+            self.ts.event_client_finished.set()
+
+            return IperfOutput(
```
wizhaoredhat, line 74: "IperfOutput" doesn't look right here.

thom311: I think `PerfClient` and `PerfServer` should be renamed to `TaskClient` and `TaskServer`.
And `perf.py` (the file that contains those two classes) should just merge with `task.py`.
And `IperfOutput` should be renamed to `FlowTestOutput`.

That renaming will happen in a later patch (that is not yet written). But for now, the `Evaluator` operates on `flow_test IperfOutput`. Despite the naming, this is not Iperf-specific...

wizhaoredhat: Ok thanks! Glad you noticed this too

PR #78, opened 2024-07-05
[th/serialize-output] add methods for serializing/parsing the TftAggregateOutput

thom311:
We write the result to a YAML and parse them back.

This mostly uses `common.dataclass_{to,from}_dict()`, except that the list of `TftAggregateOutput` is (at the highest level) not a dataclass (it's a dictionary with a"tft-tests" key and a list of `TftAggregateOutput`). That fact makes the parsing more effort than if it were just a dataclass that we can pass to `common.dataclass_from_dict()`.

Also, use `common.dataclass_to_dict()` in `Evaluator.dump_to_json()`

### DIFF
--- a/evaluator.py
+++ b/evaluator.py
@@ -3,7 +3,6 @@ import json
 import sys
 import yaml
 
-from dataclasses import asdict
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Optional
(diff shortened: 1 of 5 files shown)


wizhaoredhat:
LGTM

PR #81, opened 2024-07-08
[th/oc-run-exec] task: use Task.run_oc_exec() instead of Task.run_oc()

thom311:
Note that in some cases we would pass on the command to the PluginOutput. So this changes the output. However, I think the command in the output is only for informational purpose, so this is probably fine. You anyway couldn't copy+paste the command, because the string already previously lacked the "--kubeconfig" and "-n" options, that are automatically added by Task.run_oc().


wizhaoredhat:
LGTM

PR #82, opened 2024-07-08
[th/eval-result-rework] cleanup evaluation of results

thom311:
- extract parsing of iperf and netperf output to (unit tested) helper functions
- let the parsing don't fail with an exception, instead signal parsing errors in the result (`success=False`).
- let the test already compute the `Bitrate` and embed it in the result. This moves the functionality from the `Evaluator` right to the place where we collect the test result. It also means, the result JSON contains also the bit rate.
- implement bitrate and evaluation for `netperf` test type.

### DIFF
--- a/common.py
+++ b/common.py
@@ -219,6 +219,19 @@ def enum_convert_list(enum_type: Type[E], value: Any) -> list[E]:
     return output
 
 
+T = TypeVar("T")
+
+
+def dict_get_typed(d: typing.Mapping[Any, Any], key: Any, vtype: type[T]) -> T:
+    try:
+        v = d[key]
+    except KeyError:
+        raise KeyError(f'missing key "{key}"')
+    if not isinstance(v, vtype):
+        raise TypeError(f'key "{key}" expected type {vtype} but has value "{v}"')
+    return v
+
+
 def j2_render(in_file_name: str, out_file_name: str, kwargs: dict[str, Any]) -> str:
     with open(in_file_name) as inFile:
         contents = inFile.read()
@@ -240,9 +253,6 @@ def serialize_enum(
         return [serialize_enum(item) for item in data]
     else:
         return data
-
-
-T = TypeVar("T")
 
 
 def dataclass_to_dict(obj: "DataclassInstance") -> dict[str, Any]:
--- a/evaluator.py
+++ b/evaluator.py
@@ -8,7 +8,6 @@ from pathlib import Path
 from typing import Optional
 
 import evalConfig
-import pluginbase
 import tftbase
 
 from common import dataclass_to_dict
@@ -108,8 +107,7 @@ class Evaluator:
             result = self._eval_flow_test(run.flow_test)
             test_results.append(result)
             for plugin_output in run.plugins:
-                plugin = pluginbase.get_by_name(plugin_output.name)
-                plugin_result = plugin.eval_log(
+                plugin_result = plugin_output.plugin.eval_log(
                     plugin_output, run.flow_test.tft_metadata
                 )
                 if plugin_result is not None:
--- a/evaluator.py
+++ b/evaluator.py
@@ -84,7 +84,7 @@ class Evaluator:
             test_id=md.test_case_id,
             test_type=md.test_type,
             reverse=md.reverse,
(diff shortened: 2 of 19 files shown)


common.py
```diff
@@ -557,3 +557,49 @@ def ethtool_stat_get_packets(data: dict[str, str], packet_type: str) -> Optional
     if not any_match:
         return None
     return total_packets
+
+
+def netperf_parse(testname: Any, data: str) -> dict[str, float]:
+
+    # We want to accept both the testname as string or as TestType.  Note that
```
wizhaoredhat, line 564: Why does testname have 2 possible types?

tests/test_common.py
```diff
@@ -492,3 +492,72 @@ def test_ethtool_parse_stat() -> None:
...
+    assert common.netperf_parse("TCP_STREAM", data) == {
+        "Receive Socket Size Bytes": 131072.0,
+        "Send Socket Size Bytes": 16384.0,
+        "Send Message Size Bytes": 16384.0,
+        "Elapsed Time Seconds": 30.0,
+        "Throughput 10^6bits/sec": 35710.68,
+    }
+
+    with pytest.raises(TypeError):
+        common.netperf_parse("xxx", data)
```
wizhaoredhat, line 552: What is "xxx"?

common.py
thom311: this is in `common.py`, but `TestType` is in `tftbase.py`. `common.py` cannot use `tftbase.py`.

For the most part, `testname` is just a string and that's it.

However, the callers actually have a `TestType.NETPERF_*` enum at hand. Allow them to pass that enum directly. `common.netperf_parse()` doesn't really know about that enum, but it doesn't care. It only cares that `str(testname)` is one of the expected arguments.

tests/test_common.py
thom311: an invalid argument `testname` for the unit test, that results in a `TypeError` exception.

common.py
thom311: Alternatives:

- move the code to `tftbase.py` and have `testname: str | TestType` (or only `testname: TestType`). I didn't do that, because this code is for the most part independent from ocp-t-f-t (hence in `common.py`). Also, `TestType` contains other values like `HTTP`, so accepting `TestType` is not great.

- only requiest a string. Then the caller must somehow convert the `TestType` that they have to the right string.

The current alternative is chosen, because it's convenient for the caller (they can just pass their `TestType` enum), while the parsing of the input `testname` is still quite trivial (convert to string, and compare that in a relaxed way).

tests/test_common.py
thom311: changed `"xxx"` to `"bogus-name"`

common.py
wizhaoredhat: It is kind of unexpected to see this code in common.py since it is specific for netperf. Could we move the code to related netperf code?

thom311: The only place that has netperf related code is `testTypeNetPerf.py`. That module depends on other ocp-t-f-t code. `testTypeNetPerf.py` is not simple, because you need to understand how it integrates into ocp-t-f-t.

Moving code away from `testTypeNetPerf.py` makes that `testTypeNetPerf.py` a bit simpler -- which is the most important aspect, because `testTypeNetPerf.py` is the complicated part. It also encourages to make the code self-contained, unit tested and handle edge cases. Which it could have been also while residing inside `testTypeNetPerf.py`, but it wasn't (point in case!).

There is also `common.ethtool_stat_parse()`. If we dislike this, then we should add `ethtool_utils.py`, `netperf_utils.py` (or leave them in `pluginValidateOffload.py` and `testTypeNetPerf.py`?). I don't see value in splitting up `common.py` (or keeping functionality out of it), as long as all the pieces are self-contained, Python-stdlib-only, simple components, and do a clear thing. Unlike most other modules, `common.py`/`tftbase.py` does not require that it's functions share a common topic/purpose.

thom311: Despite what I said. The code now moved away from `common.py` and branch rebased. Better?

wizhaoredhat:
LGTM, thanks @thom311 !

common.py
wizhaoredhat: Looks good!

PR #84, opened 2024-07-12
[th/abort-on-exception] task: abort process on uncaught exception from Task's thread

thom311:
Exceptions from threads are caught by the Thread class and returned during join().

Unexpected exceptions are always bugs. We want to notice them early and tear down the process. No need to wait for the Thread.join(), which also causes the log output to be filled with messages from other threads, and the relevant exception log is easy to miss.

While at it, also crash the main thread (and the entire process) on timeout during Thread.join(). That is a bug too.

### DIFF
--- a/task.py
+++ b/task.py
@@ -1,5 +1,6 @@
 import enum
 import json
+import os
 import shlex
 import sys
 import threading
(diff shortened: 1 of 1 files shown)


SalDaniele:
LGTM

PR #85, opened 2024-07-12
[th/task-aggregate-output] task: let base implementation aggregate_output() handle flow-test result

thom311:
Instead of having TestTypeHandler implementations override _aggregate_output(), move the common code to Task.aggregate_output().

If a TestType implementation really wants to overwrite Task.aggregate_output(), it still can. However, that should not be necessary, because Task.aggregate_output() already all that should be necessary. In particular, it honors IperfOutput's "success" field, which allows Task.aggregate_output()  to do the right thing, without having a per-test-type override.

Task._aggregate_output() still exists and some implementations use it for logging. That part is still task specific, because the base class does not know how to interpret the details of the task.

### DIFF
--- a/task.py
+++ b/task.py
@@ -1,5 +1,6 @@
 import enum
 import json
+import logging
 import os
 import shlex
 import sys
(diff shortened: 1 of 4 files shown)


task.py
```diff
@@ -435,7 +436,28 @@ def aggregate_output(self, out: tftbase.TftAggregateOutput) -> None:
...
+            logger.log(log_level, f"Results of {self.ts.get_test_str()}: {log_msg}")
+            logger.debug(f"result: {common.dataclass_to_json(result)}")
+
+            if type(self)._aggregate_output is Task._aggregate_output:
+                # This instance did not overwrite _aggregate_output(). This is
+                # fine for a task that returned the IperfOutput. Don't call
+                # _aggregate_output.
+                return
+
+        self._aggregate_output(result, out)
```
SalDaniele, line 460: Wouldn't this fail now since we removed this method from the classes derived from Task?

thom311: It's correct. See the lines above.

We have tests (`testType*py`) and plugins (`plugin*.py`). The former return `IperfOutput`, the latter return `PluginOutput` (both are `AggregatableOutput`).

Tasks that return an `AggregatableOutput` (either `IperfOutput` or `PluginOutput`) MUST overwrite `_aggreate_output()`, ... unless the task is a test (which returns `IperfOutput`), then it may choose not overwrite `_aggregate_output()`. The reason is to provide the convenience to not add code for the case where the test-task has nothing to aggregate further.

thom311: extended code comment about that.

thom311: @SalDaniele to be clear. This "convenience" to allow the test to not implement `_aggregate_output()` comes at the cost of requiring a few lines of code and several lines of code comments explaining why it's done that way. Additionally, surprising the reviewer.

Arguably, one might ask, wouldn't it be simpler to just require every implementation to just implement an empty `_aggregare_output()`? Maybe. But note that in the future we might have more test implementations, at which point this convenience starts to matter more.

This is not a very convincing argument. So... yeah. I just did it that way, for not very good reasons. It could also be different...

PR #86, opened 2024-07-12
[th/eval-all-plugins] evaluator: enable plugin evaluation for all plugins

thom311:
Of the 3 existing plugins, only "validate_offload" implemented Plugin.eval_log(), which is called by Evaluator for collecting the result. All our plugins should be collected in the results.

Instead of implementing eval_log() for all plugins, rework how it works.

The existing eval_log() implementation of PluginValidateOffload only performed some static check on the PluginOutput. Now, already when running the test, this check is run and PluginOutput.success is set.

Later, the default implementation of Plugin.eval_plugin_output() (previously Plugin.eval_log()) considers PluginOutput.success. This way, we have it implemented for all plugin types, they just need to set the right PluginOutput.success.

Note that Plugin.eval_plugin_output() no longer does anything plugin-specific. It only considers PluginOutput.success. In the future, we may want that the valuation needs to consider the plugin, so we keep this as a virtual function of Plugin.

### DIFF
--- a/evaluator.py
+++ b/evaluator.py
@@ -104,8 +104,9 @@ class Evaluator:
             result = self._eval_flow_test(run.flow_test)
             test_results.append(result)
             for plugin_output in run.plugins:
-                plugin_result = plugin_output.plugin.eval_log(
-                    plugin_output, run.flow_test.tft_metadata
+                plugin_result = plugin_output.plugin.eval_plugin_output(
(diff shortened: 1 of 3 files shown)


SalDaniele:
LGTM, just leaving a note that it would be good to add additional plugin information (i.e. message, specific results) to the pluginOutput. I imagine this will come in later PRS

PR #87, opened 2024-07-12
[th/test-meta-data-indexes] TestMetadata: record the tft/test_cases/connections indexes

thom311:
The input configuration for TFT is a YAML with nested lists of tests, test-cases and connections. Record in the test meta data of the output the indexes, so that we know for which particular test the respective result is.

This allows to associate the test result with the corresponding entry in the test config YAML.


SalDaniele:
LGTM

PR #88, opened 2024-07-12
[th/refactor-eval-result] evaluator: refactor duplicate code for the evaluation result

thom311:
With `evaluator.py` we can evaluate the test result. Note that `main.py` also can do evaluation right after running the test. There is thus duplicated functionality.

Move some of the duplicated code to Evaluator class.


SalDaniele:
LGTM

PR #90, opened 2024-07-15
[th/plugin-result-name] plugin: record the plugin name in PluginResult

thom311:
Otherwise it is pretty hard to see, which of the results came from which plugin.


SalDaniele:
Thanks LGTM. Might also be worth adding specific metrics, i.e. in the case of validate offload the ratio of offloaded to not-offloaded packets (similar to how we provide bitrate in the perf test output).

PR #91, opened 2024-07-15
[th/validate-offload-fix-parsing] fix parsing of "validate_offload" result

thom311:
validate_offload calls `ethtool -S`, but we might only get `rx` results.

The commit 6e0e4f622188 added some error checking, and would always expect a `tx` result too. Fix that.


SalDaniele:
LGTM

Issue #25, opened 2024-04-30
We need a more reliable way to determine "Cluster Type"


SalDaniele:
Currently, we simply inspect the name of the kubeconfig files that exist to determine what sort of cluster we are deploying in (NIC mode versus DPU mode etc).

We should have something more robust, just because a tenant cluster config exists doesn't necessarily mean we are using a DPU mode cluster.

PR #92, opened 2024-07-15
[th/config-kubeconfig] make kubeconfig/kubeconfig_infra configurable

thom311:
Make kubeconfig/kubeconfig_infra configurable in the config YAML.

Based on that, we also detect the cluster mode SINGLE/DPU.

Fixes: https://github.com/wizhaoredhat/ocp-traffic-flow-tests/issues/25

The autodetection still works as before. As far as autodetect goes, I think that is good enough. Anything further requires explicit configuration (which this PR adds).

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,8 @@ tft:
         plugins:
           - name: (15)
           - name: (15)
+kubeconfig: (16)
+kubeconfig_infra: (16)
 ```
 
 1. "name" - This is the name of the test. Any string value to identify the test.
@@ -94,6 +96,9 @@ tft:
     | measure_cpu      | Measure CPU Usage    |
     | measure_power    | Measure Power Usage  |
     | validate_offload | Verify OvS Offload   |
+16. "kubeconfig", "kubeconfig_infra": if set to non-empty strings, then these are the KUBECONFIG
+  files. "kubeconfig_infra" must be set for DPU cluster mode. If both are empty, the configs
+  are detected based on the files we find at /root/kubeconfig.*.
 
 ## Running the tests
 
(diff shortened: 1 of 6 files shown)


thom311:
@SalDaniele the questions is, whether you had any other "autodetection" in mind. This doesn't really detect anything, but it allows the user/caller to explicitly select the kubeconfigs (and indirectly the ClusterMode).

SalDaniele:
I didn't have anything specific in mind. It just seemed like relying on the kubeconfig name could be risky, the user might not know to use a specific name for their kubeconfig as the knob to configure the flow tests.

We could check the actual OpenShift objects, like the sriov node policy or NAD to better understand how the cluster is configured as well.

I think for now Kubeconfig is fine though, given that we are not using the two-cluster infra/tenant design at the moment anyways. We can re-consider ways to add smarter autodetection when adding dpu operator testing to this tool.

wizhaoredhat:
@SalDaniele Is there something wrong with having it explicitly defined?

SalDaniele:
Nope, having it explicitly defined would be better in my opinion than relying on the naming of the kubeconfig

SalDaniele:
LGTM

thom311:
> We could check the actual OpenShift objects, like the sriov node policy or NAD to better understand how the cluster is configured as well.

Except, the configuration to detect is the KUBECONFIG itself. Before you have that, it seems hard to check for Openshift objects.

I guess, we could somehow honor the KUBECONFIG from the environment. Maybe with that we could ask the cluster whether there is a DPU somewhere. But it seems impossible from there to find the kubeconfig_infra for the DPU. We could detect this way `ClusterMode.SINGLE`, and error out, if we detect that this cluster has a DPU (and require the user to explicitly set `kubeconfig`+`kubeconfig_infra` in the config).

TODO. Maybe.

PR #93, opened 2024-07-15
[th/test-metadata-in-result] record the TestMetadata in the result

thom311:
The TestMetadata structure describes which test we run. Embed that one in the Result, instead of picking a few properties.


wizhaoredhat:
LGTM. Better to consolidate the attributes into 1 obj.

PR #94, opened 2024-07-16
[th/simple-tcp-client-server-test] simple: add a test type with a simple python tcp echo server/client

thom311:
The purpose is to have an alternative test tool. In this case, it's a relatively simple Python script where the client connects via TCP, sends a message, and the server sends it back.

The point is also, that this script is relatively easily extendable, either for manual testing or to add a new mode. Hence you can specify additional command line arguments in the test configuration.

https://issues.redhat.com/browse/NHE-743

### DIFF
--- a/scripts/simple-tcp-server-client.py
+++ b/scripts/simple-tcp-server-client.py
@@ -1,0 +1,307 @@
+#!/bin/python3
+
+import argparse
+import os
+import random
+import socket
+import sys
+import time
+
+from contextlib import contextmanager
+from typing import Iterator
+from typing import Optional
+from typing import Callable
+
+
+DEFAULT_ADDR = "127.0.0.1"
+DEFAULT_BUFSIZE = 65495
+DEFAULT_DURATION = 0.0
+DEFAULT_NUM_CLIENTS = 1
+DEFAULT_PORT = 5201
+DEFAULT_SLEEP = 0.001
+
+global_start_time = time.monotonic()
+
+
+def create_socket() -> socket.socket:
+    return socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+
+@contextmanager
+def socket_timeout(
+    log_name: str,
+    s: socket.socket,
+    start_time: float,
+    duration: float,
+    *,
+    done_msg: Optional[Callable[[], None]] = None,
+) -> Iterator[socket.socket]:
+    if duration <= 0.0:
+        t = None
+    else:
+        t = (start_time + duration) - time.monotonic()
+        if t <= 0.0:
+            print(f"{log_name}: duration expired. Quit")
+            sys.exit(0)
+    s.settimeout(t)
+
+    try:
+        yield s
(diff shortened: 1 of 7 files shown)


scripts/simple-tcp-server-client.py
```diff
@@ -0,0 +1,307 @@
+#!/bin/python3
```
wizhaoredhat, line 1: Are we sure there is not like a iperf like python library we can use? Or do you think writing the code from a low level has some advantages?

thom311: I think the main point of this PR is to add all the basics for a python-based tool.

It works as is, but probably does not something very useful.

The use is to have some code in place, that can be modified and adjusted when we have a particular use.

Maybe an iperf-like library would be useful, and then the script should be easily extendable to support that, once we have the exact use-case.

wizhaoredhat: Ah ok thanks!

testConfig.py
```diff
@@ -277,6 +347,7 @@ def parse(
             server=tuple(server),
             client=tuple(client),
             plugins=tuple(plugins),
+            simple_type=simple_type,
```
wizhaoredhat, line 350: I am a little bit confused why we need an extra state variable here.

Does this represent like arguments to a command? It might be useful to have this for other test types as well. I image maybe we need to add an extra arg to iperf for certain unique test scenarios.

thom311: > I am a little bit confused why we need an extra state variable here.

Need is a strong word.

I thought, the use case for the simple script, is that you patch it for some testing (or future enhancements). When doing that, for testing you will necessarily need to build the test image and set `TFT_TEST_IMAGE` (to not use   `quay.io/wizhao/tft-tools:latest`). At that point, it seems maybe useful to pass additional arguments to the script.

The `simple_type` key contains test-type specific parameters for `TestType.SIMPLE`. In practice, those are extra command line arguments that are blindly passed on. Granted, that might also be useful for `TestType.IPERF_*`. But for that case, I thought we would add a `iperf_type` key (alongside `simple_type`).

Maybe extra command line arguments makes sense for all test types (in some form). So maybe the `simple_type.client_args` and `simple_type.server_args` should be hoisted to the top-level, and in some sense apply to various test types. Unclear which is preferable... (?)

wizhaoredhat: Which way do you think would be better? I am ok with hoisting it to the top-level. There was extra arg support in the old bash traffic flow test. It would definitely be nice if we prepare for something like that (not necessarily coding it)

thom311: for arguments like `client_args`/`server_args`, that is, in an abstract sense some extra arguments we pass on (e.g. to iperf), I think it makes sense to hoist them to the higher level. That will replace `simple_type` entry.

I will do that...

wizhaoredhat: Thanks!

wizhaoredhat:
LGTM

PR #95, opened 2024-07-16
[th/container-fix-shutdown] fix shutdown of test pods, use tini init process and minor cleanups

thom311:
- `oc delete` is used to cleanup the created pods. Those pods ran `/sbin/init`, which was unkillable by SIGTERM. Hence, every test would waste 30 seconds waiting to kill those containers. Replace that with `tini`, which can be killed fast.

See also: https://github.com/krallin/tini

- enable EPEL/CBR in the test image `quay.io/wizhao/tft-tools:latest`. Also install `psproc-ng` and `coreutils` for better usability of the container when debugging.

### DIFF
--- a/images/Containerfile
+++ b/images/Containerfile
@@ -11,6 +11,8 @@ RUN curl -L -o /etc/yum.repos.d/benchmark:openSUSE_Factory.repo https://download
 RUN dnf install -y epel-next-release epel-release
 
 RUN dnf install \
+        --allowerasing \
+        coreutils \
         cri-tools \
         ethtool \
         git \
--- a/images/Containerfile
+++ b/images/Containerfile
@@ -25,6 +25,7 @@ RUN dnf install \
         net-tools \
         netperf \
         pciutils \
+        procps-ng \
         python3 \
         sysstat \
         tcpdump \
--- a/images/Containerfile
+++ b/images/Containerfile
@@ -31,6 +31,7 @@ RUN dnf install \
         python3 \
         sysstat \
         tcpdump \
+        tini \
         util-linux \
         vim \
         wget \
@@ -38,4 +39,7 @@ RUN dnf install \
 
 RUN mkdir -p /etc/ocp-traffic-flow-tests && echo "ocp-traffic-flow-tests" > /etc/ocp-traffic-flow-tests/data
 
-CMD ["/bin/bash"]
+COPY ./images/container-entry-point.sh /usr/bin/container-entry-point.sh
+
+ENTRYPOINT ["/usr/bin/container-entry-point.sh"]
+CMD ["/usr/bin/sleep", "infinity"]
--- a/images/Containerfile
+++ b/images/Containerfile
@@ -1,8 +1,14 @@
 FROM quay.io/centos/centos:stream9
 
+RUN dnf install -y 'dnf-command(config-manager)'
+
+RUN dnf config-manager --set-enabled crb
+
 RUN curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable.repo https://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/CentOS_9_Stream/devel:kubic:libcontainers:stable.repo
 
 RUN curl -L -o /etc/yum.repos.d/benchmark:openSUSE_Factory.repo https://download.opensuse.org/repositories/benchmark/openSUSE_Factory/benchmark.repo
+
+RUN dnf install -y epel-next-release epel-release
 
 RUN dnf install \
         cri-tools \
--- a/images/Containerfile
+++ b/images/Containerfile
@@ -4,7 +4,28 @@ RUN curl -L -o /etc/yum.repos.d/devel:kubic:libcontainers:stable.repo https://do
 
 RUN curl -L -o /etc/yum.repos.d/benchmark:openSUSE_Factory.repo https://download.opensuse.org/repositories/benchmark/openSUSE_Factory/benchmark.repo
 
-RUN INSTALL_PKGS="vim wget jq python3 git cri-tools net-tools iptables iproute pciutils ethtool httpd iperf3 tcpdump sysstat ipmitool util-linux netperf nc iputils" && yum install -y ${INSTALL_PKGS}
+RUN dnf install \
+        cri-tools \
+        ethtool \
+        git \
+        httpd \
+        iperf3 \
+        ipmitool \
+        iproute \
+        iptables \
+        iputils \
+        jq \
+        nc \
+        net-tools \
+        netperf \
+        pciutils \
+        python3 \
+        sysstat \
+        tcpdump \
+        util-linux \
+        vim \
+        wget \
+        -y
 
 RUN mkdir -p /etc/ocp-traffic-flow-tests && echo "ocp-traffic-flow-tests" > /etc/ocp-traffic-flow-tests/data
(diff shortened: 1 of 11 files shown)


images/Containerfile
```diff
@@ -1,11 +1,45 @@
...
+        wget \
+        -y
 
 RUN mkdir -p /etc/ocp-traffic-flow-tests && echo "ocp-traffic-flow-tests" > /etc/ocp-traffic-flow-tests/data
 
-CMD ["/bin/bash"]
+COPY ./images/container-entry-point.sh /usr/bin/container-entry-point.sh
+
+ENTRYPOINT ["/usr/bin/container-entry-point.sh"]
+CMD ["/usr/bin/sleep", "infinity"]
```
wizhaoredhat, line 45: Why is there entry point and CMD?

thom311: CMD is there, so that you can run `podman run --rm --name foo quay.io/wizhao/tft-tools:latest` without specifying a command.

The CMD is separate from ENTRYPOINT, so that theoretically you could run `podman run --rm --name foo quay.io/wizhao/tft-tools:latest $CMD`, though, I am not sure whether that is really useful.

When we create the pod in openshift, it anyway overwrites both ENTRYPOINT and CMD and doesn't use it.

task.py
```diff
@@ -265,13 +265,13 @@ def get_namespace(self) -> str:
...
 
-    def get_template_args(self) -> dict[str, str]:
+    def get_template_args(self) -> dict[str, str | list[str]]:
         return {
             "name_space": self.get_namespace(),
             "test_image": tftbase.get_tft_test_image(),
             "image_pull_policy": tftbase.get_tft_image_pull_policy(),
-            "command": "/sbin/init",
-            "args": "",
+            "command": "/usr/bin/container-entry-point.sh",
```
wizhaoredhat, line 273: Looks a bit weird calling the entry point script as a command here...

thom311: why is  that weird?

To my understanding, the `ENTRYPOINT`+`CMD` is the way of how the image is called by default. For example, when you run `podman run --rm --name foo quay.io/wizhao/tft-tools:latest` (that is, without specifying `--entrypoint` and not providing a `CMD` argument).

In the pod CR's `"command"` and `"args"` complete replace the `ENTRYPOINT`/`CMD`, as they fully specify command is called.

In this case, the pod's `"command"` is the same as `ENTRYPOINT`.

Note, that the pod's `"args"` is left empty. So this invocation is not exactly `ENTRYPOINT`+`CMD` but only `ENTRYPOINT`. If we wanted it to be the same, we had to set `"args": ["sleep", "infinity"]`. That is not exactly done, but note `container-entry-point.sh` defaults to `sleep infinity` when called without arguments, so we get the same result. This is done, so we don't redundantly have to set `"args"` and `container-entry-point.sh` alone knows what we want to do.

wizhaoredhat:
Sorry let me read up more on entrypoint and commands in docker files.

images/Containerfile
wizhaoredhat: Sorry why does it override both ENTRYPOINT and CMD in openshift? I am pretty sure it still honors it in openshift.

task.py
wizhaoredhat: I am confused because the purpose of entrypoint is to not allow any arbitrary command to go through. But in the cases here we can, even when the entrypoint is specified in the container. 
Does the execution look like this:
`usr/bin/container-entry-point.sh /usr/bin/container-entry-point.sh` (i think you can run ps -aux on the worker)
First one is from entry point in docker file
Second one is from the command above?

thom311: > I am confused because the purpose of entrypoint is to not allow any arbitrary command to go through.

ENTRYPOINT is used in some cases as how to run the container (`podman run`). But it's not used in others (`podman run` or the way we run the container in openshift).

You can see this yourself:


```
$ podman build --no-cache -t foo -f images/Containerfile .
```

Then:

```
$ time podman run --rm --name x foo &
[1] 14355
$ podman exec -ti x ps -aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.3  0.0   2648  1280 ?        Ss   07:04   0:00 /usr/bin/tini -p SIGTERM -g -e 143 -- /usr/bin/sleep infinity
root           2  0.0  0.0   2652  1408 ?        S    07:04   0:00 /usr/bin/sleep infinity
root           3  0.0  0.0  16412  5888 pts/0    Rs+  07:04   0:00 ps -aux
$ time podman stop x
x

real    0m0.187s
user    0m0.095s
sys     0m0.039s
$ 
real    0m12.492s
user    0m0.138s
sys     0m0.106s
[1]+  Done                    time podman run --rm --name x foo
```
```
$ time podman run --rm --name x foo sleep 1000 &
[1] 14494
$ podman exec -ti x ps -aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.3  0.0   2648  1152 ?        Ss   07:05   0:00 /usr/bin/tini -p SIGTERM -g -e 143 -- sleep 1000
root           2  0.0  0.0   2652  1408 ?        S    07:05   0:00 sleep 1000
root           3  0.0  0.0  16412  6016 pts/0    Rs+  07:05   0:00 ps -aux
$ time podman stop x
x

real    0m0.182s
user    0m0.091s
sys     0m0.037s
$ 
real    0m12.322s
user    0m0.145s
sys     0m0.087s
[1]+  Done                    time podman run --rm --name x foo sleep 1000
```

Note:
- PID1 is `tini`, because  `container-entry-point.sh` script calls `exec tini`.
- `podman stop` works. `podman kill -s TERM` would also work. And so does `oc delete ...` succeed to terminate the pod fast.
- ENTRYPOINT+CMD is used during `podman run` if no CMD is provided.
- ENTRYPOINT+`$args` is used during `podman run ... $args`
- `podman exec` does not use ENTRYPOINT nor CMD
- as PIDs in above example, we have
  - PID1 for `tini` (via  `exec` in `container-entry-point.sh`
  - PID2, the sleep command started by `tini`
  - PID3, the `ps` from `podman exec`. This didn't go through `container-entry-point.sh`, otherwise the PID would be larger.
- the use of `tini` here is similar to what you would get with `podman|docker run --init`. Except, that only works with podman/docker, while explicitly calling `tini` works with any container runtime (crictl called by kubernetes).

Finally, read https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#define-a-command-and-arguments-when-you-create-a-pod :

> The command field corresponds to ENTRYPOINT, and the args field corresponds to CMD in some container runtimes.

The effects of

```
    command:
      - "/usr/bin/container-entry-point.sh"
     args:
       - sleep
       - infinity
```
```
    command:
      - "/usr/bin/container-entry-point.sh"
     args: []
```
```
    command: []
     args: []
```
are basically all the same. The branch chooses the second variant, because partly is explicit (by setting `command`/`ENTRYPOINT`) and partly relies on the default behavior of running "/usr/bin/container-entry-point.sh" without arguments.


This branch is correct.

wizhaoredhat: Ah I see `The command field corresponds to ENTRYPOINT, and the args field corresponds to CMD in some container runtimes.`

I learnt something important here. Thanks @thom311 . I appreciate you explaining this to me.

wizhaoredhat:
LGTM

PR #96, opened 2024-07-16
[th/measure-power-parse] pluginMeasurePower: improve error reporting of "measure_power" plugin

thom311:
There is no need for individual logger.error() calls. We already call self.run_oc_exec(), which will log an ERROR with the full output of the command. That is more useful already.

Also, when running ipmitool or parsing its output fails, signal an error and set a reason message.

### DIFF
--- a/pluginMeasurePower.py
+++ b/pluginMeasurePower.py
@@ -1,7 +1,10 @@
 import re
 import time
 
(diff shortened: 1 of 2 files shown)


wizhaoredhat:
LGTM

PR #97, opened 2024-07-18
Added tests for multi-home and multi-network

csahithi:
This MR adds three new tests to test multi-homing and multi-networkpolicies scenarios:
- POD_TO_POD_2ND_INTERFACE_SAME_NODE
- POD_TO_POD_2ND_INTERFACE_DIFF_NODE
- POD_TO_POD_MULTI_NETWORK_POLICY

These tests assume that a secondary network already exists. A sample network attachment definition to test the changes is given below:
```
apiVersion: k8s.cni.cncf.io/v1
kind: NetworkAttachmentDefinition
metadata:
  annotations:
    meta.helm.sh/release-name: ovn-kubernetes
    meta.helm.sh/release-namespace: default
  name: ovn-secondary
  namespace: default
  labels:
    tft-tests: "{{ index }}"
spec:
  config: '{ "cniVersion": "0.4.0", "name": "ovn-secondary", "netAttachDefName": "default/ovn-secondary",
    "isSecondary": true, "subnets": "10.245.0.0/16/24", "mtu": 1400, "type": "ovn-k8s-cni-overlay",
    "topology":"layer3", "logFile": "/var/log/ovn-kubernetes/ovn-k8s-cni-overlay.log",
    "logLevel": "5", "logfile-maxsize": 100, "logfile-maxbackups": 5, "logfile-maxage":
    5 }'
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,8 +42,9 @@ tft:
         plugins:
           - name: (15)
           - name: (15)
-kubeconfig: (16)
-kubeconfig_infra: (16)
+        secondary_network_nad: "(16)"
+kubeconfig: (17)
+kubeconfig_infra: (17)
 ```
 
 1. "name" - This is the name of the test. Any string value to identify the test.
@@ -79,6 +80,9 @@ kubeconfig_infra: (16)
     | 24 | HOST_TO_NODE_PORT_TO_HOST_DIFF_NODE |
     | 25 | POD_TO_EXTERNAL |
     | 26 | HOST_TO_EXTERNAL |
+    | 27 | POD_TO_POD_2ND_INTERFACE_SAME_NODE |
+    | 28 | POD_TO_POD_2ND_INTERFACE_DIFF_NODE |
+    | 29 | POD_TO_POD_MULTI_NETWORK_POLICY |
 4. "duration" - The duration that each individual test will run for.
 5. "name" - This is the connection name. Any string value to identify the connection.
 6. "type" - Supported types of connections are iperf-tcp, iperf-udp, netperf-tcp-stream, netperf-tcp-rr
@@ -96,7 +100,8 @@ kubeconfig_infra: (16)
     | measure_cpu      | Measure CPU Usage    |
     | measure_power    | Measure Power Usage  |
     | validate_offload | Verify OvS Offload   |
-16. "kubeconfig", "kubeconfig_infra": if set to non-empty strings, then these are the KUBECONFIG
+16. "secondary_network_nad" - The name of the secondary network for multi-homing and multi-networkpolicies tests.
+17. "kubeconfig", "kubeconfig_infra": if set to non-empty strings, then these are the KUBECONFIG
   files. "kubeconfig_infra" must be set for DPU cluster mode. If both are empty, the configs
   are detected based on the files we find at /root/kubeconfig.*.
 
--- a/config.yaml
+++ b/config.yaml
@@ -22,5 +22,7 @@ tft:
           - name: measure_cpu
           - name: measure_power
           - name: validate_offload
+        # Secondary network is required for tests 27, 28 and 29
+        # secondary_network_nad: "default/ocp-secondary"
 kubeconfig:
 kubeconfig_infra:
--- a/task.py
+++ b/task.py
@@ -282,6 +282,7 @@ class Task(ABC):
             "args": [],
             "index": f"{self.index}",
             "node_name": self.node_name,
+            "secondary_network_nad": self.ts.connection.secondary_network_nad,
         }
 
     def render_file(
@@ -383,6 +384,20 @@ class Task(ABC):
             raise RuntimeError("Failure to get static.podIP for {self.pod_name}")
         return pod_ip
 
+    def get_secondary_ip(self) -> str:
+        jsonpath = "{.metadata.annotations.k8s\\.ovn\\.org\\/pod-networks}"
+        r = self.run_oc(
+            f"get pod {self.pod_name} -o jsonpath='{jsonpath}'", die_on_error=True
+        )
+
+        y = yaml.safe_load(r.out)
+        ip_address_with_cidr = typing.cast(
+            str, y[self.ts.connection.secondary_network_nad]["ip_address"]
+        )
+        ip_address = ip_address_with_cidr.split("/")[0] if ip_address_with_cidr else ""
+        logger.info(f"Secondary IP: {ip_address}")
+        return ip_address
+
     def create_cluster_ip_service(self) -> str:
         in_file_template = "./manifests/svc-cluster-ip.yaml.j2"
         out_file_yaml = "./manifests/yamls/svc-cluster-ip.yaml"
@@ -418,6 +433,56 @@ class Task(ABC):
 
         return self.run_oc(
             "get service tft-nodeport-service -o=jsonpath='{.spec.clusterIP}'"
+        ).out
+
+    def create_ingress_multi_network_policy(self, ingressPort: int) -> str:
+        in_file_template = "./manifests/allow-ingress-mnp.yaml.j2"
+        out_file_yaml = "./manifests/yamls/allow-ingress-mnp.yaml"
+
+        template_args = {
+            **self.get_template_args(),
+            "ingress_port": f"{ingressPort}",
+        }
+
+        self.render_file(
+            "Ingress Multi Network Policy",
+            in_file_template,
+            out_file_yaml,
+            template_args,
+        )
+        r = self.run_oc(f"apply -f {out_file_yaml}", may_fail=True)
+        if r.returncode != 0:
+            if "already exists" not in r.err:
+                logger.info(r)
+                sys.exit(-1)
+
+        return self.run_oc(
+            "get multi-networkpolicies allow-ingress-mnp", die_on_error=True
+        ).out
+
+    def create_egress_multi_network_policy(self, egressPort: int) -> str:
+        in_file_template = "./manifests/allow-egress-mnp.yaml.j2"
+        out_file_yaml = "./manifests/yamls/allow-egress-mnp.yaml"
+
+        template_args = {
+            **self.get_template_args(),
+            "egress_port": f"{egressPort}",
+        }
+
+        self.render_file(
+            "Egress Multi Network Policy",
+            in_file_template,
+            out_file_yaml,
+            template_args,
+        )
+        r = self.run_oc(f"apply -f {out_file_yaml}", may_fail=True)
+        if r.returncode != 0:
+            if "already exists" not in r.err:
+                logger.info(r)
+                sys.exit(-1)
+
+        return self.run_oc(
+            "get multi-networkpolicies allow-egress-mnp", die_on_error=True
         ).out
 
     def start_setup(self) -> None:
@@ -589,6 +654,15 @@ class ServerTask(Task, ABC):
             in_file_template = ""
             out_file_yaml = ""
             pod_name = EXTERNAL_PERF_SERVER
+        elif connection_mode in (
+            ConnectionMode.MULTI_HOME,
+            ConnectionMode.MULTI_NETWORK,
+        ):
+            in_file_template = "./manifests/pod-secondary-network.yaml.j2"
+            out_file_yaml = (
+                f"./manifests/yamls/pod-secondary-network-{node_name}-server.yaml"
+            )
+            pod_name = f"normal-pod-secondary-network-{node_name}-server-{port}"
         elif pod_type == PodType.SRIOV:
             in_file_template = "./manifests/sriov-pod.yaml.j2"
             out_file_yaml = f"./manifests/yamls/sriov-pod-{node_name}-server.yaml"
@@ -635,6 +709,10 @@ class ServerTask(Task, ABC):
 
             self.cluster_ip_addr = self.create_cluster_ip_service()
             self.nodeport_ip_addr = self.create_node_port_service(self.port + 25000)
+
+        if self.connection_mode == ConnectionMode.MULTI_NETWORK:
+            self.create_ingress_multi_network_policy(self.port)
+            self.create_egress_multi_network_policy(self.port)
 
     def confirm_server_alive(self) -> None:
         if self.connection_mode == ConnectionMode.EXTERNAL_IP:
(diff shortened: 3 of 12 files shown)


csahithi:
@wizhaoredhat PTAL

manifests/host-pod.yaml.j2
```diff
@@ -14,9 +14,9 @@ spec:
   containers:
   - name: {{ pod_name }}
     image: {{ test_image }}
-    command:
-      - "{{ command }}"
-    args: {{ args }}
+    command: {{ command }}
```
wizhaoredhat, line 17: Does this fix an issue? Or just a style change?

manifests/pod.yaml.j2
```diff
@@ -6,13 +6,15 @@ metadata:
   labels:
     tft-tests: "{{ index }}"
     tft-port: "{{ port }}"
+  annotations:
+    k8s.v1.cni.cncf.io/networks: {{ secondary_network_nad }}
```
wizhaoredhat, line 10: What if secondary_network_nad doesn't exist? Shouldn't this be optionally included?

task.py
```diff
@@ -265,13 +265,18 @@ def get_namespace(self) -> str:
...
+        return self.ts.cfg_descr.get_tft().oci_bin
+    
     def get_template_args(self) -> dict[str, str]:
         return {
             "name_space": self.get_namespace(),
             "test_image": tftbase.get_tft_test_image(),
             "image_pull_policy": tftbase.get_tft_image_pull_policy(),
-            "command": "/sbin/init",
-            "args": "",
+            # /sbin/init is not working with containerd. The pod goes into CrashLoopBackOff state
```
wizhaoredhat, line 276: Could you try with  https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/95

trafficFlowTests.py
```diff
@@ -129,6 +141,7 @@ def _run_test_case_instance(
         for t in servers + clients + monitors:
             t.initialize()
 
+        time.sleep(10)
```
wizhaoredhat, line 144: Why is there a 10 second sleep here?

wizhaoredhat (commented):
I would like @thom311 to also take a look.

Thanks for the PR!

trafficFlowTests.py
thom311: I also think that the sleep should be avoided.

It's a bit ugly that the `initialize()` part does not run in parallel (which maybe should be fixed). So creating the test pods takes already longer than necessary. But adding another sleep on top of that, increases the time waiting for the result.

task.py
thom311: I also think #95 is the solution.

This runs `/sbin/init` (that is, systemd) as non-PID 1. While there are cases where that is desirable (`--user`), I think it's wrong here.

README.md
```diff
@@ -42,8 +42,10 @@ tft:
         plugins:
           - name: (15)
           - name: (15)
-kubeconfig: (16)
-kubeconfig_infra: (16)
+        secondary_network_nad: "(16)"
+    oci_bin: (17)
+kubeconfig: (18)
+kubeconfig_infra: (18)
```
thom311, line 48: I think, `oci_bin` as a configuration parameter makes sense. But maybe it should be at the highest level of configuration (beside the `kubeconfig` entry). Do you expect to run tests with docker and some with podman, during the same run?

thom311: maybe you could split this branch. A first commit that introduces `oci_bin`. That is pretty independent from the rest. It makes the individual commits smaller and easier to review. It also gives room for (multiple) commit messages that talk about the change.

config.yaml
```diff
@@ -22,5 +23,6 @@ tft:
           - name: measure_cpu
           - name: measure_power
           - name: validate_offload
+        secondary_network_nad: "default/ovn-secondary"
```
thom311, line 26: this `config.yaml` is an example, but also maybe something that a user takes as a start for creating their configuration.

The `secondary_network_nand` option and the `oci_bin: docker`) settings don't seem what a user would do by default. Maybe comment these lines out, but leave them here (as comments) to show to the user that those options exist.

task.py
```diff
@@ -342,6 +347,19 @@ def get_pod_ip(self) -> str:
...
         y = yaml.safe_load(r.out)
         return typing.cast(str, y["status"]["podIP"])
+    
+    def get_secondary_ip(self) -> str:
+        jsonpath = '{.metadata.annotations.k8s\.ovn\.org\/pod-networks}'
+        logger.info(f"get pod {self.pod_name} -o jsonpath='{jsonpath}'")        
+        r = self.run_oc(f"get pod {self.pod_name} -o jsonpath='{jsonpath}'")
+        if r.returncode != 0:
+            logger.info(r)
+            sys.exit(-1)
```
thom311, line 357: drop the `if r.returncode != 0` block and pass `die_on_error=True` to `self.run_oc()`.

task.py
```diff
@@ -342,6 +347,19 @@ def get_pod_ip(self) -> str:
         r = self.run_oc(f"get pod {self.pod_name} -o yaml", die_on_error=True)
         y = yaml.safe_load(r.out)
         return typing.cast(str, y["status"]["podIP"])
+    
+    def get_secondary_ip(self) -> str:
+        jsonpath = '{.metadata.annotations.k8s\.ovn\.org\/pod-networks}'
+        logger.info(f"get pod {self.pod_name} -o jsonpath='{jsonpath}'")
```
thom311, line 353: not sure about this logging line. Note that `self.run_oc()` already does DEBUG logging with more information (in particular, the result of the command). 

At INFO level, it would be better to just log the secondary IP that was received at the end of this function, instead of "we are about to request some information from the pod, but don't log the actual result".

perf.py
```diff
@@ -217,13 +228,19 @@ def get_target_ip(self) -> str:
             external_pod_ip = self.get_podman_ip(self.server.pod_name)
             logger.debug(f"get_target_ip() External connection to {external_pod_ip}")
             return external_pod_ip
+        elif self.connection_mode in (ConnectionMode.MULTI_NETWORK, ConnectionMode.MULTI_HOME):
```
thom311, line 231: there are various lines in the PR that would be reformatter by `python-black`. Please run python-black on it.

Also note the github action, which checks for this style. Basically, fix issues in the "Check style, test / build" github test.

manifests/pod.yaml.j2
thom311: and maybe, `ConfConnection.secondary_network_nad` should be an `Optional[str]` to represent missing as `None` (instead of `""`).

Granted, we can interpret `""` as missing. To me, it feels cleaner to represent that as `None`.

task.py
wizhaoredhat: #95 Was merged. Could you fix the conflicting files and see if it fixes it for you.

thom311:
@csahithi there were quite a few changes on `main` recently. If that causes problems during rebasing the branch, I am sorry. But now the work there should slow down.

If you could please rebase once more onto latest `main`?

trafficFlowTests.py
```diff
@@ -48,6 +48,14 @@ def _cleanup_previous_testspace(self, cfg_descr: ConfigDescriptor) -> None:
             logger.error(r)
             raise Exception("cleanup_previous_testspace(): Failed to delete services")
         logger.info(f"Cleaned services with label tft-tests in namespace {namespace}")
+        logger.info(f"Cleaning multi-networkpolicies with label tft-tests in namespace {namespace}")
+        r = cfg_descr.tc.client_tenant.oc(
+            f"delete multi-networkpolicies -n {namespace} -l tft-tests"
+        )
+        if r.returncode != 0:
+            logger.error(r)
+            raise Exception("cleanup_previous_testspace(): Failed to delete multi-networkpolicies")
```
thom311, line 57: let's not do additional error handling. Just pass `die_on_error=True` to `K8sClient.oc()` to abort on failure.

Or, maybe don't treat such error fatally. `K8sClient.oc()` already does by default `may_fail=False`, which means a failure results in a warning log line. That may be sufficient for as far as handling the error goes.

thom311: The reason is that the line automatically logged by `K8sClient.oc()` is better than `logger.error(r)`, which lacks any context about what command failed. So, you can do less work (lines of code) while getting the better logging line.

csahithi: @thom311 Sure! I'll make these changes. Also I will rebase my branch. 
I've marked this PR as a draft for now so that I can mark it as ready once I've completed all the changes and tested them.

manifests/allow-egress-veth-mnp.yaml.j2
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: k8s.cni.cncf.io/v1beta1
+metadata:
+  namespace: {{ name_space }}
+  name: allow-egress-veth-mnp
+  labels:
+    tft-tests: "{{ index }}"
+  annotations:
+    k8s.v1.cni.cncf.io/policy-for: {{ secondary_network_nad }}
+spec:
+  podSelector: {}
```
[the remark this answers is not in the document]
csahithi: Yes, podSelector need not be defined for this

manifests/host-pod.yaml.j2
csahithi: This change is not required anymore after merging with #95

manifests/pod.yaml.j2
csahithi: Thanks for the suggestion! Modified the PR to include `secondary_network_nad` optionally

task.py
csahithi: Merging with #95 fixed this. Thanks!

README.md
csahithi: Sure! I've removed the `oci_bin` changes from this PR. Will raise a PR for those changes separately after this PR is merged

perf.py
csahithi: Reformatted using python-black

trafficFlowTests.py
csahithi: Added this initially, since EXTERNAL_IP tests were failing because of delay in docker container bring up. But the tests are now working with the latest changes and the latest image. So, removed this timeout now

csahithi:
@thom311 @wizhaoredhat Pls review the changes. Thanks!

thom311:
The `Check style, test / build` test in github still fails.

You can also manually run
```
$ flake8
$ mypy --strict --config-file mypy.ini .
```

csahithi:
@thom311 
> The `Check style, test / build` test in github still fails.
> 
> You can also manually run
> 
> ```
> $ flake8
> $ mypy --strict --config-file mypy.ini .
> ```

I see that one of the failures is 
```
./task.py:389:47: W605 invalid escape sequence '\.'
./task.py:389:52: W605 invalid escape sequence '\.'
./task.py:389:57: W605 invalid escape sequence '\/'
```
this is because of `jsonpath = '{.metadata.annotations.k8s\.ovn\.org\/pod-networks}'`

This is required to run ` kubectl get pod normal-pod-secondary-network-ovn-worker-server-5201 -o jsonpath='{.metadata.annotations.k8s\.ovn\.org\/pod-networks}'`
Without the escape characters the command fails.
How can I handle this?

task.py
```diff
@@ -420,6 +436,54 @@ def create_node_port_service(self, nodeport: int) -> str:
...
+            out_file_yaml,
+            template_args,
+        )
+        r = self.run_oc(f"apply -f {out_file_yaml}", may_fail=True)
+        if r.returncode != 0:
+            if "already exists" not in r.err:
+                logger.info(r)
+                sys.exit(-1)
+
+        return self.run_oc("get multi-networkpolicies allow-ingress-mnp").out
```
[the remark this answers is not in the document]
csahithi: The reason why I added this check here instead of adding `die_on_error=True` is because if the resource already exists then it might fail even though it shouldn't

thom311:
> @thom311
> How can I handle this?

At the first layer/level, you have here a string in Python. Backslashes there have already a meaning, as you want to literally have a backslash inside the string, you must backslash-escape the backslash.

Like `"jsonpath = '{.metadata.annotations.k8s\\.ovn\\.org\\/pod-networks}'"`

task.py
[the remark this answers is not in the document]
thom311: I am talking about the `run_oc("get...")`. That one should not fail, as we just created the resource.

If you would anticipate it to fail (which you don't need too), then you also should handle errors somehow. `die_on_error=True` is a IMO suitable way to handle it here.

config.yaml
```diff
@@ -22,5 +22,7 @@ tft:
           - name: measure_cpu
           - name: measure_power
           - name: validate_offload
+        # Secondary network is required for tests 27, 28 and 29
```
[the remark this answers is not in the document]
csahithi: @thom311 `POD_TO_POD_2ND_INTERFACE_SAME_NODE`, `POD_TO_POD_2ND_INTERFACE_DIFF_NODE` and `POD_TO_POD_MULTI_NETWORK_POLICY` are of type `TestCaseType` which is one level below the `TestType`. So this check I believe cannot be applied here. 
I have raised a ValueError in case `secondary_network_nad `is None in `get_secondary_ip()` since this method gets called only for these three TestCaseTypes.
```
def get_secondary_ip(self) -> str:
        if self.ts.connection.secondary_network_nad == None:
            raise ValueError(f'Secondary Network is missing')
```

Do you think this is good enough?

thom311:
the unit tests fail. I think you can just to another commit first, to delete the tests 
- `test_test_case_type_to_connection_mode()`
- `test_test_case_type_is_same_node()`
- `test_test_case_type_to_server_pod_type()`
- `test_test_case_type_to_client_pod_type()`

Those are no longer useful,and you don't need to adjust them for the new `TestCaseType`s.

config.yaml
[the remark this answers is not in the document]
csahithi: why is `namespace` argument being added to the `parse `method for this fix?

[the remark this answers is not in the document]
csahithi: Thanks for the suggestion! I modified it accordingly

task.py
```diff
@@ -612,6 +688,10 @@ def __init__(self, ts: TestSettings):
         self.out_file_yaml = out_file_yaml
         self.pod_name = pod_name
 
+        if self.connection_mode == ConnectionMode.MULTI_NETWORK:
+            self.create_ingress_multi_network_policy(self.port)
+            self.create_egress_multi_network_policy(self.port)
```
thom311, line 693: it's important that the `__init__()` does not already perform such actions that have side effects. That is, because the subclass probably will chain up the `super().__init__(...)` method. Since `ServerTask.__init__()` performs relevant initializations, the subclass probably want to chain up as first thing. That leaves not a nice place how the subclass can still affect what the parent's `__init__()` does.

The simple solution is to move this code to `initialize()`. That's what the method is there for.

task.py
```diff
@@ -383,6 +384,22 @@ def get_pod_ip(self) -> str:
             raise RuntimeError("Failure to get static.podIP for {self.pod_name}")
         return pod_ip
 
+    def get_secondary_ip(self) -> str:
+        if self.ts.connection.secondary_network_nad is None:
+            raise ValueError("Secondary Network is missing")
```
thom311, line 389: this check can be dropped now. The value is always set (possibly to `f"{namespace}/ocp-secondary"`.

thom311 (commented):
rest lgtm. Thank you.

thom311 (approved):

PR #100, opened 2024-07-22
[th/jinja2-command] fix/rework setting the "command" in Jinja2 manifest templates

thom311:
- fix "args" jinja2 variable for "persistent" mode.

- make the "{{ command }}" a plain list via `command: {{ command }}` instead of `command: [ "{{ command }}" ]`. The code that generates the template variables is now in full control of specifying a list, instead of something inside a list inside double quotes.

- in "persistent" mode, let tasks only overwrite "args" (CMD) instead of "command" (ENTRYPOINT). That way, they inherit the entrypoint of the Task parent process, which is sensibly set to `/usr/bin/container-entry-point.sh`.

### DIFF
--- a/manifests/host-pod.yaml.j2
+++ b/manifests/host-pod.yaml.j2
@@ -14,8 +14,7 @@ spec:
   containers:
   - name: {{ pod_name }}
     image: {{ test_image }}
-    command:
-      - "{{ command }}"
+    command: {{ command }}
     args: {{ args }}
     imagePullPolicy: {{ image_pull_policy }}
     securityContext:
--- a/manifests/pod.yaml.j2
+++ b/manifests/pod.yaml.j2
@@ -12,7 +12,6 @@ spec:
   containers:
   - name: {{ pod_name }}
     image: {{ test_image }}
-    command:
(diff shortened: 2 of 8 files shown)


manifests/host-pod.yaml.j2
```diff
@@ -14,8 +14,7 @@ spec:
   containers:
   - name: {{ pod_name }}
     image: {{ test_image }}
-    command:
-      - "{{ command }}"
+    command: {{ command }}
```
SalDaniele, line 17: Shouldnt the ENTRYPOINT be a single string? Any additional args would be provided as an {{ arg }} IIUC

thom311: As far as Kubernetes is concerned, the "command" (akin to the ENTRYPOINT) and the "args" (akin to the CMD) are both lists.

In practice, we only pass the command name there, so our list will only have one entry. But it's a list still.
So for us, it would also work to keep it as it was. I find it a bit nicer, to have this however a list in the template.

SalDaniele:
LGTM

PR #101, opened 2024-07-22
[th/jinja2-render-data] common: add common.j2_render_data() and add test

thom311:
The behavior of Jinja2 is not obvious. Add a unit test for exposing the behavior.

Also, move the jinja2 import inside the j2_render_data() function, the
only place where it's used. "common.py" generally works with the
standard lib only, and does not depend on additional pip packages.
Having jinja2 there is a bit ugly. Move it, so you can still fully use
common.py without jinja2, except the functions j2_render() and
j2_render_data().

### DIFF
--- a/common.py
+++ b/common.py
@@ -1,7 +1,6 @@
 import abc
 import collections
 import dataclasses
-import jinja2
 import json
 import typing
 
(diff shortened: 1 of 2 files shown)


wizhaoredhat:
LGTM

PR #102, opened 2024-07-22
[th/python3-in-test-image] install python3.11 venv and basic python sources in test container image

thom311:
In   `quay.io/wizhao/tft-tools:latest`:

- install python3.11 and setup a venv in `/usr/opt/pyvenv-3.11` with a few pip package.
- copy some of the base python sources to `/usr/opt/ocp-tft`. Currently, this is not very exciting, it will be more interesting when we have a few python utilities there, that we can run via the container (e.g. via `oc exec`).

### DIFF
--- a/images/Containerfile
+++ b/images/Containerfile
@@ -49,6 +49,12 @@ RUN /opt/pyvenv3.11/bin/python -m pip install \
         kubernetes \
         pytest
 
+COPY \
+    common.py \
+    host.py \
+    logger.py \
+    /opt/ocp-tft/
+
 RUN mkdir -p /etc/ocp-traffic-flow-tests && echo "ocp-traffic-flow-tests" > /etc/ocp-traffic-flow-tests/data
 
 COPY ./images/container-entry-point.sh /usr/bin/container-entry-point.sh
--- a/images/Containerfile
+++ b/images/Containerfile
@@ -12,6 +12,7 @@ RUN dnf install -y epel-next-release epel-release
 
 RUN dnf install \
         --allowerasing \
+        /usr/bin/python \
         coreutils \
         cri-tools \
         ethtool \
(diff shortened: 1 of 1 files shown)


images/Containerfile
```diff
@@ -29,6 +30,7 @@ RUN dnf install \
         pciutils \
         procps-ng \
         python3 \
```
wizhaoredhat, line 32: Can we remove python3 then?

images/Containerfile
```diff
@@ -12,6 +12,7 @@ RUN dnf install -y epel-next-release epel-release
 
 RUN dnf install \
         --allowerasing \
+        /usr/bin/python \
```
wizhaoredhat, line 15: This looks out of place...?

images/Containerfile
thom311: no. `dnf` depends on `python3`. We cannot uninstall that (it's probably in the container to begin with). Even if it were possible, it seems a bad idea to drop the "system" python (that is, the version that is deemed to be suitable for all other packaged python3 programs of the distro).

images/Containerfile
thom311: This is there to have the "unversioned" `/usr/bin/python` executable. This is just a symlink to `./python3`. It is not there normally, but is provided by the package `python-unversioned-command`. I install it, so that a plain `python` command also works.

With `dnf install`, it's IMO perfectly fine to specify the file you want to install (`/usr/bin/python`), instead of the package name `python-unversioned-command`. I did it in this case, because it seams clearer what this is.

wizhaoredhat: ah i got it, it was weird to see it like this

images/Containerfile
wizhaoredhat: ok

wizhaoredhat:
LGTM

PR #103, opened 2024-07-24
[th/http-with-duration] http: honor test "duration" for HTTP test

thom311:
Previously, the test would just fetch the page once and quit (fast). It's not clear how the test duration makes sense for the HTTP test, but I think it makes the most sense to poll the page for the requested duration (with a sleep time of 0.2 seconds).


wizhaoredhat:
LGTM

PR #104, opened 2024-07-24
[th/k8sclient-kubeconfig-from-env] k8sClient: use kubeconfig from environment with K8sClient

thom311:
This is mainly useful for manual testing, when firing up K8sClient in the REPL or with `python -c "import k8sClient; ..."`.

All code in the project, still passes an explicit kubeconfig argument. And in those cases, that is probably a good idea.

### DIFF
--- a/k8sClient.py
+++ b/k8sClient.py
@@ -9,7 +9,13 @@ import host
 
 
 class K8sClient:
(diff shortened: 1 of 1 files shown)


SalDaniele:
LGTM

PR #105, opened 2024-07-24
[th/k8sClient-debug] add K8sClient.oc_{debug,get}() helpers and no longer use crictl from test image

thom311:
- add helpers

   - `K8sClient.oc_debug()` + `K8sClient.oc_get()` (which wrap `kubectl debug|get` via `K8sClient.oc()`)
   - `Task.run_oc_debug()` + `Task.run_oc_get()`

- `kubectl debug` is more cumbersome to use than `oc debug`. But those issues are mostly worked around.

- no longer use and install `crictl` (`cri-tools` package) in the test image at   `quay.io/wizhao/tft-tools:latest`. Instead, use the crictl tool from the worker node (and do so via `Task.run_oc_debug()`). 

  - to install `crictl` in `quay.io/wizhao/tft-tools:latest` we had to enable an openSUSE repository (which yesterday was 503 Service Unavailable, it's back again).
  - the test container image targets different Openshift/Kubernetes version. While it probably works, it's not clear that installing some `crictl` version will work on the different versions that run on the node. What would be better, is to use the `crictl` version that is installed on the node, that we want to interact with.

### DIFF
--- a/k8sClient.py
+++ b/k8sClient.py
@@ -1,13 +1,17 @@
+import json
 import kubernetes  # type: ignore
 import logging
 import os
 import shlex
+import sys
 import typing
 import yaml
 
 from collections.abc import Iterable
 
 import host
+
+from logger import logger
 
 
 class K8sClient:
@@ -45,6 +49,25 @@ class K8sClient:
             return shlex.split(cmd)
         return list(cmd)
 
+    def _oc_get_full_cmd(
+        self,
+        *,
+        cmd: str | Iterable[str],
+        namespace: typing.Optional[str] = None,
+    ) -> list[str]:
+        namespace_args: tuple[str, ...]
+        if namespace:
+            namespace_args = ("-n", namespace)
+        else:
+            namespace_args = ()
+        return [
+            "kubectl",
+            "--kubeconfig",
+            self._kc,
+            *namespace_args,
+            *self._oc_get_cmd(cmd),
+        ]
+
     def oc(
         self,
         cmd: str | Iterable[str],
@@ -53,19 +76,8 @@ class K8sClient:
         die_on_error: bool = False,
         namespace: typing.Optional[str] = None,
     ) -> host.Result:
-        namespace_args: tuple[str, ...]
-        if namespace:
-            namespace_args = ("-n", namespace)
-        else:
-            namespace_args = ()
         return host.local.run(
-            [
-                "kubectl",
-                "--kubeconfig",
-                self._kc,
-                *namespace_args,
-                *self._oc_get_cmd(cmd),
-            ],
+            self._oc_get_full_cmd(cmd=cmd, namespace=namespace),
             die_on_error=die_on_error,
             log_level_fail=logging.DEBUG if may_fail else logging.ERROR,
         )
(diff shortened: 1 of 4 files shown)


k8sClient.py
```diff
@@ -31,19 +38,170 @@ def get_nodes_with_label(self, label_selector: str) -> list[str]:
...
+        cmd: str | Iterable[str],
+        *,
+        node_name: str,
+        test_image: str,
+        namespace: str,
+        may_fail: bool = False,
+        die_on_error: bool = False,
+    ) -> host.Result:
+        container_name = f"ocp-tft-debug-{node_name}-{random.randint(0, 2**64-1):016x}"
+        # We use `kubectl debug` and not `oc debug`. There are thus some differences.
```
wizhaoredhat, line 147: I don't understand why we can't use the tools pod to do this by chrooting on to the host? 

Could this possibly work https://www.redhat.com/sysadmin/how-oc-debug-works ?

wizhaoredhat: But if there is an advantage of using kubectl debug, we can keep it as is.

But I am wondering if we can elevate the tools pod container to a level where it can run crictl properly.

thom311: What I initially wanted to do, is get rid of crictl in `quay.io/wizhao/tft-tools:latest` image (for reasons mentioned the PR description).

I actually only later realized, that it would have been sufficient to change
```diff
diff --git i/pluginValidateOffload.py w/pluginValidateOffload.py
index c885cf34027d..1b8b427dd8a4 100644
--- i/pluginValidateOffload.py
+++ w/pluginValidateOffload.py
@@ -171,7 +171,7 @@ class TaskValidateOffload(PluginTask):
 
     def extract_vf_rep(self) -> Optional[str]:
         r = self.run_oc_exec(
-            f"crictl --runtime-endpoint=unix:///host/run/crio/crio.sock ps -a --name={self.perf_pod_name} -o json"
+            f"/host/usr/bin/crictl --runtime-endpoint=unix:///host/run/crio/crio.sock ps -a --name={self.perf_pod_name} -o json"
         )
 
         iface: Optional[str] = None
```
So yes, we can "use the tools pod to do this by chrooting on to the host" (or without CAP_SYS_CHROOT run the binary directly from `/host/usr/bin`).

---

But I didn't understand that at first, and wanted  to use `kubectl debug`. So I added the helpers like `K8sClient.oc_{get,exec,debug}()` and their `Task.run_oc_{get,exec,debug}()` counterparts.


> But I am wondering if we can elevate the tools pod container to a level where it can run crictl properly.

We could. `kubectl|oc debug` also just runs a pod. We could implement this in various ways, by running one pod or the other. `K8sClient.oc_debug()` does all the steps of creating and deleting the container in one and requires no additional setup. Granted, the only (current) user `PluginValidateOffset.extract_vf_rep()` already has a suitable pod at hand, so there is little to no benefit.

---

With all that I know now, I would not write `K8sClient.oc_debug()`. But as it is already written, I think it's a quite a useful function to have in the toolbox, as is has fewer requirements to the caller (that is, it doesn't require the caller to start/delete a suitable pod). What's also nice, is you see the `kubectl debug` command line in the log, which you can copy+paste to reproduce manually.

thom311: ok. The problem is that this is just too slow. `oc delete` takes several seconds... (What the hell).

I close this PR, and:
- will use the host container we already for `kubectl exec ... chroot /host critctl ...`.
- reopen another PR with what I think are useful utility functions here.

thom311:
Ok. The problem is that this is just too slow. oc delete takes several seconds... (What the hell).

I close this PR, and:

    will use the host container we already for kubectl exec ... chroot /host critctl ....
    reopen another PR with what I think are useful utility functions here.

PR #108, opened 2024-07-25
[th/crictl-from-node] don't install crictl/cri-tools in test image and use the one from the node (chroot /host)

thom311:
This replaces parts of https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/105

See https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/105#issuecomment-2250160362


wizhaoredhat:
LGTM

PR #109, opened 2024-07-25
[th/k8sclient-utils] add `K8sClient.oc_{get,exec}()` helpers 

thom311:
- add `K8sClient.oc_exec()`. That already existed as `Task.run_oc_exec()`, but now the core functionality moved from `Task` to `K8sClient`. The former just calls the latter (with specifying some default arguments like the namespace)

- add `K8sClient.oc_get()` and `Task.run_oc_get()`. This seems useful, as it also parses the requested JSON. Parsing JSON is easy, what is useful here is that it handles errors (by logging or `die_on_error`).

This is a re-spin of https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/105


task.py
```diff
@@ -300,23 +300,33 @@ def render_file(
     def initialize(self) -> None:
         pass
 
+    @property
+    def client(self) -> K8sClient:
+        return self.tc.client(tenant=self.tenant)
+
+    def _run_oc_namespace(
```
wizhaoredhat, line 307: We aren't really running anything. Could we give this a better name?
I.e. get_namespace_from_args

k8sClient.py
```diff
@@ -37,19 +43,97 @@ def get_nodes_with_label(self, label_selector: str) -> list[str]:
             for e in self._client.list_node(label_selector=label_selector).items
         ]
 
+    @staticmethod
+    def _oc_get_cmd(cmd: str | Iterable[str]) -> list[str]:
```
wizhaoredhat, line 47: These names are kind of confusing since "get" is a special keyword for k8s.

thom311: renamed.

wizhaoredhat:
LGTM

PR #110, opened 2024-07-25
[th/netdev-utils] add "netdev.py" with utilities for reading sysfs and getting information from iproute2/ethtool

thom311:
This is quite extensive.

- add `netdev.ip_{links,addrs,routes}()` which wraps `ip link|addrs|routes` and parses (and validates) the output. They return Python data classes that are strongly typed.

- same for `ethtool -s`. Add `netdev.ethtool_driver()`.

- add various helpers for read sysfs files.

- since all names in kernel (e.g. interface names in `ip link` or path names in `/sys/class/net`) are not enforced to be in any encoding, the code handles that correctly by using `errors="surrogateescape"`. Almost everything should just work. Of course, the best advise is to stick to always ASCII or at least UTF-8.

- add `netdev.get_device_infos()`, which collects information about all interfaces (pciaddresses, iplink output, ethtool, their VF/VF_REP status) and returns them in a data structure, that can be printed in JSON. There are also helper functions to parse that JSON back.

- there is a main function which accepts commands. Currently, only `get_device_infos` command is supported, which collects all device informations and prints them in JSON.

- as `netdev.py` is embedded in the `quay.io/wizhao/tft-tools:latest` image, we can also call

```
$ podman run --rm --network=host quay.io/thaller/tft-tools:latest ocp-tft-netdev get_device_infos
```

or via `oc exec`.

This gives something like
```json
[
  {
    "pciaddr": "0000:ca:00.0",
    "uplink_representor": "ens7f0np0"
  },
  {
    "ifname": "1bbeefd0712_040",
    "ifindex": 40,
    "link": {
      "flags": [
        "BROADCAST",
        "MULTICAST",
        "UP",
        "LOWER_UP"
      ],
      "mtu": 1400,
      "operstate": "UP",
      "link_info_kind": null
    },
    "pciaddr": "0000:ca:00.3",
    "uplink_representor": "ens7f0np0",
    "ethtool": {
      "driver": "mlx5_core",
      "version": "5.14.0-284.75.1.el9_2.x86_64",
      "firmware_version": "16.35.3006 (MT_0000000013)"
    },
    "is_vf": {
      "uplink_representor": "ens7f0np0",
      "uplink_rep_pciaddr": "0000:ca:00.0",
      "vfindex": 1,
      "vf_rep_ifname": "ens7f0np0_1",
      "vf_rep_ifindex": 23
    }
  },
  {
    "ifname": "2e200a9662c6016",
    "ifindex": 53,
...
```

---

The point of this will be to fix finding the VF_REP for "validate_offload" plugin (https://issues.redhat.com/browse/NHE-1154). The code will first run `oc exec POD -- oc exec ocp-tft-netdev get_device_infos` to find the PCI address of `eth0`. Then it will collect `get_device_infos` in the main netns, and search the JSON for the interface that is the VR_REP for the PCI address.

### DIFF
--- a/common.py
+++ b/common.py
@@ -109,6 +109,12 @@ def iter_get_first(
     return v0
 
 
+def iter_filter_none(lst: typing.Iterable[Optional[T]]) -> typing.Iterable[T]:
+    for v in lst:
+        if v is not None:
+            yield v
+
+
 def enum_convert(
     enum_type: Type[E],
     value: Any,
--- a/common.py
+++ b/common.py
@@ -1,6 +1,7 @@
 import abc
 import collections
 import dataclasses
+import functools
 import json
 import typing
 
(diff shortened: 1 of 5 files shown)


wizhaoredhat:
LGTM

PR #112, opened 2024-07-26
[th/validate-offload-vf-rep] fix detecting VF_REP for validate_offload plugin on 4.15

thom311:
validate_offload plugin needs to detect the VF_REP interface.

On Openshift 4.14, that works by parsing the podSandboxId. This no longer works on 4.15+.

Fix that, by instead reading the information form sysfs.

https://issues.redhat.com/browse/NHE-1154

### DIFF
--- a/pluginValidateOffload.py
+++ b/pluginValidateOffload.py
@@ -170,12 +170,14 @@ class TaskValidateOffload(PluginTask):
 
     def _create_task_operation(self) -> TaskOperation:
         def _thread_action() -> BaseOutput:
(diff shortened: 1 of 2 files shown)


thom311:
btw, with this change, I noticed that one test fails on 4.15 (because the threshold is surpassed). I looked at it, it might be an unrelated problem, and I think the branch is correct. Hence, I would merge this, even if it can break some tests.

SalDaniele:
LGTM, thank you for fixing this!

PR #113, opened 2024-07-26
[th/print-results] rewrite print_results script Python

thom311:
- add `tftbase.TestResultCollection` dataclass, this allows to automatically deserialize the RESULT output (and to load it in print_results.sh)
- rewrite `print_results.sh` in Python. There is no change in behavior except:
  - running `print_results.sh` in the terminal caused `jq` to colorize the output. There is no colorization in the python script.
  - the name of the script changes from `*.sh` to `*.py`. This requires adjustment of the CI that calls the script.

@SalDaniele


SalDaniele:
LGTM, will merge once I make the change in the CI

PR #114, opened 2024-07-27
[th/jinja2util] common,jinja2util: move common.j2_render() to new "jinja2util.py" module

thom311:
"common" should not depend on additional pip modules. We should be able to copy it somewhere and use with minimal effort (e.g. not need to install additional pip modules).

This is almost the case, except for common.j2_render() and common.j2_render_data(). Move those two functions to a new module "jinja2util.py".

### DIFF
--- a/common.py
+++ b/common.py
@@ -307,23 +307,6 @@ def dict_get_typed(
     return v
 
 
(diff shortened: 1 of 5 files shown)


wizhaoredhat:
LGTM

PR #115, opened 2024-07-29
[th/fix-external-perf-server-run] fix podman-run for external-perf-server

thom311:
- don't use `podman run --init` with our recent test image `quay.io/wizhao/tft-tools:latest`, which already sets the ENTRYPOINT to `/usr/bin/container-entry-point.sh`.
- honor `TFT_IMAGE_PULL_POLICY` for `podman run`.


wizhaoredhat:
LGTM

PR #117, opened 2024-07-31
[th/k8sclient-oc-debug] add K8sClient.oc_debug() helper

thom311:
`K8sClient.oc_debug()` calls `kubectl debug` to execute a command on a node.

I initially added this for https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/105 , however, it has the downside that it needs to spin up a new pod (so the entire command takes about 2 seconds). If you already have a priviledged pod running that can access the resources you want (e.g. `/host` or the main netns), then a plain `kubectl exec` is going to be faster, and thus preferable. However, that does require you to first start a pod (and maybe clean it up later). So there is more setup involved with that approach, and it's only faster because you only need to start/delete the pod once (instead of every `kubectl debug` call).

Still, as this code was already written, I think it's useful helper to have in the `k8sClient.py` toolbox. For that reason, I'd like to merge this, even if there currently are no user of it.

This boils down to my view of `{logger,host,common,netdev}.py` not only having a purpose for the current ocp-tft code, but to be a toolbox of some python code that is useful in the larger context (e.g. this code shoud be re-used at other places -- maybe via copy+paste). If you disagree with that view, then merging this patch makes no sense, because there is no user.

I also don't want to throw away the effort of getting it right once (even if it turned out, to not use it for the purpose where I wrote it).

### DIFF
--- a/k8sClient.py
+++ b/k8sClient.py
@@ -2,6 +2,7 @@ import json
 import kubernetes  # type: ignore
 import logging
 import os
+import random
 import shlex
 import sys
 import typing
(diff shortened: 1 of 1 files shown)


SalDaniele:
I think it makes sense to park this code here for the reasons you stated.

LGTM

PR #118, opened 2024-08-02
[th/cleanup-error-logging] cleanup the error handling with host.Host.run()

thom311:
Extend `host.Host.run()` with a `check_success` callback. This allows to categorize the `Result` as success/failure, beyond that `returncode==0` check. The benefit is that `host.Host.run()` already makes an effort to provide useful logs with the result, and indicating whether the result is a success/failure will affect the error level and message. It also means, the caller does not need to invest effort to log failures, because the logging that was done automatically is already useful.

### DIFF
--- a/host.py
+++ b/host.py
@@ -301,6 +301,11 @@ class Host(ABC):
             raise decode_exception
 
         if die_on_error and not result_success:
+            import traceback
+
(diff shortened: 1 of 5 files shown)


SalDaniele:
LGTM

PR #119, opened 2024-08-02
[th/github-cleanup-install] github: cleanup pip modules in ".github/workflows/lint-test.yml"

thom311:
We already install "jc" via "requirements.txt" file.


SalDaniele:
LGTM

PR #120, opened 2024-08-05
[th/no-kubernetes-module] k8sClient: drop usage of unsued kubernetes Python library

thom311:
The kubernetes module was only used for K8sClient.get_nodes() and K8sClient.get_nodes_with_label(). Those functions were not used.

ocp-tft runs multiple threads, it's important that the lower level helpers we use are thread-safe. It's error prone to have some parts of the API not thread-safe. This could be easily fixed by adding a lock around the kubernetes calls.

Instead, K8sClient otherwise makes the choice to run the "kubectl" binary via host.Host.run(). I don't particularly care whether we access functionality via the command line or via the kubernetes Python library. But mixing the two seems unnecessary, especially for functionality that is also available via "kubectl". I think K8sClient.oc() has the benefit that we get debug logging of the command, and that we could copy+paste the command in the terminal to manually reproduce.

Also, K8sClient based on invoking kubectl could be made easily remotable by taking a host.Host argument and host.Host implementing a command based remoting (e.g. via SSH or via a podman command or via a chroot or a mix of all of those).  Currently that is not implemented (and not needed). This would not work via the kubernetes module, so we shouldn't expand on this approach.

We can still get the nodes via K8sClient.oc(). If that is too cumbersome or inconvenient, we would add a helper function that does it:
```
  $ KUBECONFIG=...
    python -c 'if 1:
        import k8sClient

        client = k8sClient.K8sClient()
        nodes = []

        nodes_output = client.oc_get("nodes")
        if nodes_output is not None:
            try:
                nodes = [item["metadata"]["name"] for item in nodes_output["items"]]
            except Exception:
                pass
        print(nodes)
    '
```


wizhaoredhat:
Good catch. LGTM

PR #121, opened 2024-08-05
[th/test-no-iproute2] netdev/tests: skip tests if iproute2 is not installed

thom311:
These tests call iproute2, the would fail if we don't have it installed. For the tests, that is not always best. Just skip the test in that case.

### DIFF
--- a/tests/test_netdev.py
+++ b/tests/test_netdev.py
@@ -1,5 +1,6 @@
-import pytest
+import functools
 import os
+import pytest
 import sys
 
 sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
It's a bit strange that iproute2 won't be installed?

thom311:
> It's a bit strange that iproute2 won't be installed?

That happens in a `podman run -ti quay.io/centos/centos bash` container (if you don't install `iproute` package). I did that to test Python3.9.

Seems something that can relatively easy happen. Ultimately, the question is whether iproute is a mandatory dependency. I think it's an optional one, so a plain `pytest` should not fail.

wizhaoredhat:
LGTM

PR #122, opened 2024-08-06
[th/move-containerfile] images: move Containerfile from images/ directory

thom311:
`podman build .` will automatically find the Containerfile if it's located in the current directory. That makes me think, that the best way for placing this file, is there.

If we would build different containers, it might make sense to place the various Containerfile in different directories. But we don't. There is only one Containerfile.

Move it.


wizhaoredhat:
LGTM

PR #126, opened 2024-08-07
[th/nhebase] add `ktoolbox` package

thom311:
EDIT: rename `nhebase` package to `ktoolbox`

the core modules `{logger,host,common,k8sClient}.py` are to a certain extend general purpose and reusable outside of ocp-traffic-flow-tests.

Move them to a new `ktoolbox` package. This directory could be copied whole to another place and used there. The package only depends on Python's stdlib (`netdev.ip_links()`, `netdev.ethernet_driver()` etc. depends on `iproute2`, `ethtool` commands, but they will fail at runtime).

Also move the tests together with the modules.

Also make those modules compatible with Python 3.9. ocp-tft as a whole requires python3.11+. That dependency makes sense, given the improvements of Python and that it's reasonably possible to get Python 3.11 (potentially via a venv). On the other hand, while RHEL9/CentOS-Stream-9 allows to install `python3.11` (and create a 3.11 venv), the default python is 3.9. So requiring 3.11 is a bit cumbersome. So adjust the `ktoolbox` module, so that it also works with Python 3.9+.


thom311:
renamed package `nhebase` to `ktoolbox`. The `k` probably stands for Kubernetes, but it's really just there to have a somewhat unique name for this package of various utils.

wizhaoredhat:
LGTM

PR #127, opened 2024-08-09
[th/host-run-sudo] host: add sudo and cwd arguments to host.Host.run()

thom311:
- support "sudo" and "cwd" arguments with the obvious meaning.

- pass to Host._run() either a shell script (as string) or a argv array. There are now cases where we call subprocess.run(..., shell=False). If the caller specifies `cmd` as string, then it's always a shell script. If the caller specifies it as an argv list, then it may or may not be executed via a shell, depending on the sudo option (but the effect should be very similar).

- catch exceptions from subprocess.run() and convert them to BinResult() failures. The way to report errors from Host.run() is not exceptions. Previously, exceptions were less likely, but now we can get them easier with a non-existing CWD or a non-existing program name in the argv list.

- be clear about when we have a shell script vs. a argv list. This patch cleans that up. If we have a str, we have a shell script. We now must know that, because combining sudo and cwd with a cmd, needs to potentially convert a script to an argv list (by prepending "/bin/sh -c") or vice versa (by shlex.join()).

- we now may log the cwd and environment variables. As always, don't use environment variables for secrets.

Examples:
```
  $ python -c 'import ktoolbox.host as h; h.local.run("echo hi", sudo=True)'
  2024-08-09 10:06:10.199 DEBUG   [th:140701195811712]: cmd[1;localhost]: call "sudo /bin/sh -c 'echo hi'"
  2024-08-09 10:06:10.214 DEBUG   [th:140701195811712]: cmd[1;localhost]: └──> "sudo /bin/sh -c 'echo hi'": success; out='hi\n'
  $ python -c 'import ktoolbox.host as h; h.local.run("echo hi", sudo=True, env={"foo":"x"})'
  2024-08-09 10:06:24.520 DEBUG   [th:140248756321152]: cmd[1;localhost]: call "sudo foo=x /bin/sh -c 'echo hi'"
  2024-08-09 10:06:24.532 DEBUG   [th:140248756321152]: cmd[1;localhost]: └──> "sudo foo=x /bin/sh -c 'echo hi'": success; out='hi\n'
  $ python -c 'import ktoolbox.host as h; h.local.run(["echo", "hi"], sudo=True, env={"foo":"x"})'
  2024-08-09 10:06:43.511 DEBUG   [th:140574963735424]: cmd[1;localhost]: call 'sudo foo=x echo hi'
  2024-08-09 10:06:43.523 DEBUG   [th:140574963735424]: cmd[1;localhost]: └──> 'sudo foo=x echo hi': success; out='hi\n'
  $ python -c 'import ktoolbox.host as h; h.local.run(["echo", "hi"], sudo=True, env={"foo":"x"}, cwd="/tmp")'
  2024-08-09 10:07:03.065 DEBUG   [th:140235035736960]: cmd[1;localhost]: call "sudo foo=x /bin/sh -c 'cd /tmp || exit 1 ; echo hi'"
  2024-08-09 10:07:03.077 DEBUG   [th:140235035736960]: cmd[1;localhost]: └──> "sudo foo=x /bin/sh -c 'cd /tmp || exit 1 ; echo hi'": success; out='hi\n'
```

### DIFF
--- a/ktoolbox/host.py
+++ b/ktoolbox/host.py
@@ -19,6 +19,9 @@ from typing import Union
 
 from .logger import logger
 
+INTERNAL_ERROR_PREFIX = "Host.run(): "
+INTERNAL_ERROR_RETURNCODE = 1
+
 
 _lock = threading.Lock()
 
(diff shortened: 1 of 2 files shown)


wizhaoredhat:
LGTM

PR #129, opened 2024-08-09
[th/firewall-nft-masquerade] firewall: add utils for nft/firewall to "ktoolbox" for masquerade

thom311:


wizhaoredhat:
Reviewed in other repo. LGTM

PR #130, opened 2024-08-20
[th/cleanup-shebang] all: use shebang "#!/usr/bin/env python3" for python scripts

thom311:
This seems best. It will honor virtual environments and pick up the right python from the path. Also be explicit about using python3, when we consider that on Fedora we don't have /usr/bin/python unless we install "python-unversioned-command" package.


wizhaoredhat:
LGTM

PR #132, opened 2024-08-20
[th/serial-expect] common: add common.Serial class for a simple minicom+pexpect

thom311:
Previously, we use minicom for opening serial ports. That was nice, because then we can use pexpect for some simple terminal interface navigation.

The problem is that minicom requires a TTY, otherwise it fails with "No cursor motion capability (cm)".

That means, if you run `podman run` without "-t"/"--tty" or `ssh HOST CMD`, you cannot easily use minicom inside the command. Maybe we could wrap minicom inside a tool that creates a tty (tmux?), but that seems cumbersome as well. Instead, use pyserial to operate the serial port.

The downside is that we no longer have a command line tool, and we cannot use pexpect anymore. Hence, implement some simple expect(). This should be sufficient for our use case, and should be easy to extend.

Add the class to ktoolbox.common. Note that the pyserial dependency is only necessary when instantiating a ktoolbox.common.Serial() class.

Also note that the class operates on str instead of bytes. But you should be able to handle any binary data with "errors='surrogatepairs'".

### DIFF
--- a/Containerfile
+++ b/Containerfile
@@ -41,6 +41,7 @@ RUN /opt/pyvenv3.11/bin/python -m pip install \
         dataclasses \
         jc \
         jinja2 \
+        pyserial \
         pytest
 RUN ln -s /opt/pyvenv3.11/bin/python /usr/bin/python-pyvenv3.11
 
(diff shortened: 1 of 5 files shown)


wizhaoredhat:
LGTM

PR #134, opened 2024-08-21
[th/host-run-cwd] fix handling `cwd` option with `host.Host.run()`

thom311:
Fix:

- don't raise an exception if we lack permission for cwd directory (instead, report a failed `Result()`)

- with `sudo=True`, change to `cwd` directory inside a shell script. That way, we evaluate the directory in the context of the sudo user, and not the calling user (who might not have permissions).


wizhaoredhat:
LGTM

PR #136, opened 2024-08-27
[th/mypy-files] mymp: set "files=." in mypy.ini

thom311:
Previously, the right way to run mypy was `mypy --strict .`. Set those
options in "mypy.ini" file. You can still override it on the command
line.


wizhaoredhat:
LGTM

PR #137, opened 2024-08-27
[th/remote-host] add host.RemoteHost for running commands via SSH/paramiko

thom311:
Add a `host.RemoteHost` class that implements the `host.Host` API (in particular, the `host.Host.run()` method). In general, the API should behave very similar whether you run `host.LocalHost().run()` or `host.RemoteHost("localhost", ...).run()`. In particular, the arguments `env`, `cwd`, `sudo` should work in a very similar manner whether you run locally or remotely.

`host.RemoteHost` is mostly stateless and immutable. It has some internal state (the paramiko client), but that is per-thread (thread-local storage) and hidden from the user.

Also add a `log_lineoutput:bool` argument to `Host.run()`. When set to True, we won't log stdout/stderr at the end, but rather line by line. That is useful for long-running commands that print their progress (for example `rsh.run("dnf upgrade -y", log_lineoutput=True)`). It is also what CDA's `host.Host._run_remote()` always does (but not `host.Host._run_local()`).

### DIFF
--- a/.github/workflows/lint-test.yml
+++ b/.github/workflows/lint-test.yml
@@ -23,6 +23,7 @@ jobs:
         python -m pip install mypy
         python -m pip install pytest
         python -m pip install types-PyYAML>=6.0.1
+        python -m pip install types-paramiko
         python -m pip install -r requirements.txt
     - name: Check code formatting with Black
       run: |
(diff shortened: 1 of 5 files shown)


wizhaoredhat:
LGTM

PR #139, opened 2024-08-28
[th/structparse-pop] common,testConfig: add helper methods for parsing simple type in testConfig

thom311:
Our input configuration is YAML, and we parse it into a `TestConfig` class. The configuration contains many simple types (`str`, `int`, `bool`, `Enum`), which should be handled in a similar fashion.

Add helper methods for handling such types and use them. The benefit is that how to handle a `str`, `int`, `bool`, `Enum` is now a common receipt instead of reimplemented at different places.

There are still complex properties that require specific code for parsing and validating them.

Note that the `common.structparse_pop_*()` methods accept a "default" argument. If the argument is missing, the parameter is mandatory. If it's given, it can be an `Optional[T]` to also return `None` for missing keys. In any case, mypy will understand depending on the "default" argument whether the function returns a `T` or `Union[None, T]`.

### DIFF
--- a/ktoolbox/common.py
+++ b/ktoolbox/common.py
@@ -1,4 +1,5 @@
 import abc
+import contextlib
 import dataclasses
 import functools
 import json
@@ -14,6 +15,7 @@ from dataclasses import fields
 from dataclasses import is_dataclass
 from enum import Enum
 from typing import Any
+from typing import Literal
 from typing import Optional
 from typing import Type
 from typing import TypeVar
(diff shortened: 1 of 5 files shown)


ktoolbox/common.py
```diff
@@ -557,30 +566,365 @@ def structparse_check_strdict(arg: Any, yamlpath: str) -> dict[str, Any]:
...
-        str, structparse_check_and_pop_name(vdict, yamlpath, required=True)
+    def for_name(self, key: str = "name") -> tuple[dict[str, Any], str, str]:
+        return self.vdict, self.yamlpath, key
+
+    def skip_check_empty(self) -> None:
+        object.__setattr__(self, "check_empty", False)
+
+
+@contextlib.contextmanager
+def structparse_with_strdict(
```
wizhaoredhat, line 591: Can we have some function headers that tell the user how to use this function as docs? This is especially helpful if we want ktoolbox to be a commonly used across projects. People need some kind of docs for these functions and their intended use.

wizhaoredhat:
LGTM!

PR #140, opened 2024-08-30
[th/unwrap] common: add unwrap() helper

thom311:
This behaves similar to Rust's unwrap() (hence the name).

It's really only useful for assertion use when we require that the value is set.

In that case, the linter might complain if we use the value as Optional[T].

This is thus both a runtime assertion and to make the linter happy, in one.

Note that the generated error message is not very good because there is no context for a better message (though, you get the backtrace, so it's not horrible). That is why it's only for assertion use.

### DIFF
--- a/ktoolbox/common.py
+++ b/ktoolbox/common.py
@@ -135,6 +135,19 @@ def iter_filter_none(lst: Iterable[Optional[T]]) -> Iterable[T]:
             yield v
 
 
+def unwrap(val: Optional[T], *, or_else: Optional[T] = None) -> T:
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
Who uses unwrap() right now?

thom311:
> Who uses unwrap() right now?

Nobody in ocp-traffic-flow-tests. It is used on CDA, for example https://github.com/thom311/cluster-deployment-automation/blob/th/clustersConfig/extraConfigDpuTenant.py#L174

wizhaoredhat:
LGTM

PR #141, opened 2024-09-02
Update traffic flow test tool to have sriov pod yaml use secondary network

vrindle:
We need to be able to test out pod2pod traffic on IPUs. Currently the dpu-operator does not deploy a webhook,so we are modifying the SRIOV pod yaml to be able to use secondary network nad and also have resource requests and limits added to the SRIOV pod if the secondary_network_nad is present.  In addition network-attachment-definition no longer by default should contain the namespace. 

In order to use this feature,

1. add `secondary_network_nad` to the yaml. 
2. `resource_name` is optional, you can set it if you would like. Otherwise the program does fetch it for us.
3. run standard pod to pod traffic on an IPU cluster. Populate config file as normal.
4. Pods will properly pass traffic through the secondary network. 

Jira issue: https://issues.redhat.com/browse/IIC-193

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -22,7 +22,10 @@ tft:
           - name: measure_cpu
           - name: measure_power
           - name: validate_offload
-        # Secondary network is required for tests 27, 28 and 29
-        # secondary_network_nad: "default/ocp-secondary"
+        # (Optional) Secondary network is required for tests 27-29. For these tests, it is mandatory; if not specified, it will default to "ocp-secondary." For other tests, 
+        # the secondary network is optional. If it is undefined, the tests will default to using the primary network.
+        # (Optional) Define the resource name for SRIOV pods, where the resource requests and limits are configured. Defaults to None if not set. If the user specified the secondary nad 
+        # it will try to autopopulate the resource name based on the nad. 
+        # resource_name: "openshift.io/dpu"
 kubeconfig:
 kubeconfig_infra:
--- a/task.py
+++ b/task.py
@@ -8,6 +8,7 @@ import threading
 import time
 import typing
 import yaml
+import functools
 
 from abc import ABC
 from abc import abstractmethod
@@ -275,6 +276,30 @@ class Task(ABC):
     def get_duration(self) -> int:
         return self.ts.cfg_descr.get_tft().duration
 
+    @functools.cache
+    @staticmethod
+    def _fetch_default_resource_name(
+        client: K8sClient, namespace: str, secondary_network_nad: Optional[str]
+    ) -> Optional[str]:
+        if secondary_network_nad is not None:
+            data = client.oc_get(
+                f"network-attachment-definition/{secondary_network_nad}",
+                namespace=namespace,
+            )
+        else:
+            data = None
+        resource_name = None
+
+        if data is not None:
+            try:
+                r = data["metadata"]["annotations"]["k8s.v1.cni.cncf.io/resourceName"]
+                if isinstance(r, str) and r:
+                    resource_name = r
+            except Exception:
+                pass
+        logger.info(f"autodetected resource_name as {repr(resource_name)}")
+        return resource_name
+
     def get_template_args(self) -> dict[str, str | list[str]]:
         return {
             "name_space": self.get_namespace(),
@@ -284,7 +309,17 @@ class Task(ABC):
             "args": [],
             "index": f"{self.index}",
             "node_name": self.node_name,
-            "secondary_network_nad": self.ts.connection.secondary_network_nad,
+            "secondary_network_nad": self.ts.connection.secondary_network_nad_or_default,
+            "use_secondary_network": (
+                "1" if self.ts.connection.secondary_network_nad else ""
+            ),
+            "resource_name": self.ts.connection.resource_name
+            or Task._fetch_default_resource_name(
+                self.client,
+                self.get_namespace(),
+                self.ts.connection.secondary_network_nad,
+            )
+            or "",
         }
 
     def render_file(
@@ -303,6 +338,7 @@ class Task(ABC):
         logger.info(
             f'Generate {log_info} "{out_file_yaml}" (from "{in_file_template}", for {self.log_name})'
         )
+
         rendered = jinja2util.j2_render(in_file_template, out_file_yaml, template_args)
 
         rendered_dict = yaml.safe_load(rendered)
@@ -381,7 +417,22 @@ class Task(ABC):
         pod_ip = None
         try:
             if y:
-                pod_ip = y["status"]["podIP"]
+                if self.ts.connection.secondary_network_nad:
+                    network_status_str = y["metadata"]["annotations"][
+                        "k8s.v1.cni.cncf.io/network-status"
+                    ]
+                    network_status = json.loads(network_status_str)
+
+                    namespace = self.get_namespace()
+                    namespace_and_nad = (
+                        f"{namespace}/{self.ts.connection.secondary_network_nad}"
+                    )
+                    for network in network_status:
+                        if network["name"] == namespace_and_nad:
+                            pod_ip = network["ips"][0]
+                            break
+                else:
+                    pod_ip = y["status"]["podIP"]
         except Exception:
             pass
         if not isinstance(pod_ip, str):
@@ -394,9 +445,13 @@ class Task(ABC):
             f"get pod {self.pod_name} -o jsonpath='{jsonpath}'", die_on_error=True
         )
 
+        namespace = self.get_namespace()
         y = yaml.safe_load(r.out)
         ip_address_with_cidr = typing.cast(
-            str, y[self.ts.connection.secondary_network_nad]["ip_address"]
+            str,
+            y[f"{namespace}/{self.ts.connection.secondary_network_nad_or_default}"][
+                "ip_address"
+            ],
         )
         ip_address = ip_address_with_cidr.split("/")[0] if ip_address_with_cidr else ""
         logger.info(f"Secondary IP: {ip_address}")
--- a/testConfig.py
+++ b/testConfig.py
@@ -197,9 +197,15 @@ class ConfConnection(StructParseBaseNamed):
     server: tuple[ConfServer, ...]
     client: tuple[ConfClient, ...]
     plugins: tuple[ConfPlugin, ...]
-    secondary_network_nad: str
+    secondary_network_nad: Optional[str]
+    resource_name: Optional[str]
 
     def serialize(self) -> dict[str, Any]:
+        extra: dict[str, Any] = {}
+        common.dict_add_optional(
+            extra, "secondary_network_nad", self.secondary_network_nad
+        )
+        common.dict_add_optional(extra, "resource_name", self.resource_name)
         return {
             **super().serialize(),
             "type": self.test_type.name,
@@ -207,8 +213,14 @@ class ConfConnection(StructParseBaseNamed):
             "server": [s.serialize() for s in self.server],
             "client": [c.serialize() for c in self.client],
             "plugins": [p.serialize() for p in self.plugins],
-            "secondary_network_nad": self.secondary_network_nad,
+            **extra,
         }
+
+    @property
+    def secondary_network_nad_or_default(self) -> str:
+        if self.secondary_network_nad is None:
+            return "ocp-secondary"
+        return self.secondary_network_nad
 
     @staticmethod
     def parse(
@@ -262,7 +274,19 @@ class ConfConnection(StructParseBaseNamed):
 
             secondary_network_nad = common.structparse_pop_str(
                 *varg.for_key("secondary_network_nad"),
-                default=f"{namespace}/ocp-secondary",
+                default=None,
+            )
+
+            if secondary_network_nad is not None and "/" in secondary_network_nad:
+                if not secondary_network_nad.startswith(f"{namespace}/"):
+                    raise ValueError(
+                        f'"{yamlpath}.secondary_network_nad": value {repr(secondary_network_nad)} must not contain a namespace. The referenced NAD is implicitly in the test namespace {repr(namespace)}'
+                    )
+                secondary_network_nad = secondary_network_nad[len(f"{namespace}/") :]
+
+            resource_name = common.structparse_pop_str(
+                *varg.for_key("resource_name"),
+                default=None,
             )
 
         if len(server) > 1:
@@ -300,6 +324,7 @@ class ConfConnection(StructParseBaseNamed):
             client=client,
             plugins=plugins,
             secondary_network_nad=secondary_network_nad,
+            resource_name=resource_name,
         )
 
 
--- a/README.md
+++ b/README.md
@@ -43,8 +43,9 @@ tft:
           - name: (15)
           - name: (15)
         secondary_network_nad: "(16)"
-kubeconfig: (17)
-kubeconfig_infra: (17)
+	resource_name: "(17)"
+kubeconfig: (18)
+kubeconfig_infra: (18)
 ```
 
 1. "name" - This is the name of the test. Any string value to identify the test.
@@ -100,8 +101,10 @@ kubeconfig_infra: (17)
     | measure_cpu      | Measure CPU Usage    |
     | measure_power    | Measure Power Usage  |
     | validate_offload | Verify OvS Offload   |
-16. "secondary_network_nad" - The name of the secondary network for multi-homing and multi-networkpolicies tests.
-17. "kubeconfig", "kubeconfig_infra": if set to non-empty strings, then these are the KUBECONFIG
+16. "secondary_network_nad" - (Optional) - The name of the secondary network for multi-homing and multi-networkpolicies tests. For tests except 27-29, the primary network will be used if unspecified (the default which is None). For mandatory tests 27-29 it defaults to "ocp-secondary" if not set.
+17. "resource_name" - (Optional) - The resource name for tests that require resource limit and requests to be set. This field is optional and will default to None if not set, but if secondary network nad is defined, traffic flow test
+tool will try to autopopulate resource_name based on the secondary+network_nad provided.
+18. "kubeconfig", "kubeconfig_infra": if set to non-empty strings, then these are the KUBECONFIG
   files. "kubeconfig_infra" must be set for DPU cluster mode. If both are empty, the configs
   are detected based on the files we find at /root/kubeconfig.*.
 
--- a/manifests/sriov-pod.yaml.j2
+++ b/manifests/sriov-pod.yaml.j2
@@ -3,8 +3,9 @@ kind: Pod
 metadata:
   namespace: {{ name_space }}
   name: {{ pod_name }}
-  annotations:
-    v1.multus-cni.io/default-network: {{ default_network }}
+  annotations: {% if not use_secondary_network %}
+    v1.multus-cni.io/default-network: {{ default_network }} {% else %}
+    k8s.v1.cni.cncf.io/networks: {{ secondary_network_nad }} {% endif %}
   labels:
     tft-tests: "{{ index }}"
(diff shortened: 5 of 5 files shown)


wizhaoredhat:
These changes will most likely not work. The tool was built with some assumptions on the primary network. Adding secondary on the yaml won't change anything with the tests (e.g. getting IP address and creating services). The configuration should have a new option to tell the tool whether it is primary or secondary network testing. If it is secondary then getting the IP addresses would need to change.

config.yaml
```diff
@@ -22,7 +22,9 @@ tft:
           - name: measure_cpu
           - name: measure_power
           - name: validate_offload
-        # Secondary network is required for tests 27, 28 and 29
+        # (Optional) Secondary network is required for tests 27, 28 and 29
```
thom311, line 25: "(Optional) ... is required" is a bit of a contradiction.

Could you slightly reword? If the parameter is omitted in the YAML, it simply defaults to `f"{namespace}/ocp-secondary"`, so it's always there.

Maybe something like:
```
# Secondary network is required for tests POD_TO_POD_2ND_INTERFACE_SAME_NODE (27), POD_TO_POD_2ND_INTERFACE_DIFF_NODE (28), POD_TO_POD_MULTI_NETWORK_POLICY (29). If unspecified, defaults to "{namespace}/ocp-secondary".
```

testConfig.py
```diff
@@ -301,6 +325,16 @@ def parse(
         else:
             secondary_network_nad = v
 
+        v = vdict.pop("resource_name", None)
+        if v is None:
+            resource_name = get_resource_name()
```
thom311, line 330: `get_resource_name()` has vast side effects, fetching the value from the current cluster.

While some attributes in `testConfig.py` are computed and don't directly come from the YAML (e.g. `ConfPlugin.plugin`), I think it's important that `TestConfig` primarily represents the content from the YAML file with only minor (stateless, side-effect-free) additions.

This means, make `resource_name: Optional[str]`, and leave it at None if it's unspecified in the YAML. Then, the caller who uses this field (`Task.get_template_args()`) has to resolve the default value, like

```python
    def get_template_args(self) -> dict[str, str | list[str]]:
        return {
            ...
            "secondary_network_nad": self.ts.connection.secondary_network_nad,
            "resource_name": self.ts.connection.resource_name or _fetch_default_resource_name(),
        }
```

---

Theoretically, we could also add a `ConfConnection.get_real_resource_name(self)->str:` method, that resolves the default value. That would be OK, since the data class still represents purely the YAML with an additional utility getter. But since there is only one caller (`Task.get_template_args()`), let `Task` handle this.

---

If you make `resource_name: Optional[str]`, then in `ConfConnection.serialize()`, maybe:
```diff
     def serialize(self) -> dict[str, Any]:
+        extra: dict[str, Any] = {}
+        common.dict_add_optional(extra, "resource_name", self.resource_name)
         return {
             **super().serialize(),
             "type": self.test_type.name,
             "instances": self.instances,
             "server": [s.serialize() for s in self.server],
             "client": [c.serialize() for c in self.client],
             "plugins": [p.serialize() for p in self.plugins],
             "secondary_network_nad": self.secondary_network_nad,
-            "resource_name": self.resource_name,
+            **extra,
         }
```

testConfig.py
```diff
@@ -36,6 +37,27 @@
...
 
+def get_resource_name() -> str:
+    try:
+        cmd = ["oc", "get", "network-attachment-definition", "-o", "yaml"]
+        nad_yaml = subprocess.check_output(cmd)
+    except subprocess.CalledProcessError as e:
+        logger.info(
+            f"Get network-attachment-defintion command failed with exit status {e.returncode}"
+        )
+        return "openshift.io/dpu"
```
thom311, line 48: Note, that ocp-tft uses `kubectl` instead of `oc`. Presumably, to work with other kubernetes variants aside openshift. Can this be fetched via kubectl?

Also, there is `host.Host.run()` which abstracts calling processes. It should be used instead of `subprocess.check_output()`, because most notably it provides useful debug logging and error handling. Further, there is `K8sClient.oc*()`, which wraps `host.Host.run()`. That latter is the one that should be used here, because it exists exactly for this purpose. The intent is that `K8sClient.oc*()` is better than what you could reasonably re-implement here (whether it succeeds at that, you are welcome to question).

IOW, I think we should call here `data = k8sClient.oc_get("network-attachment-definition")`.

If you require `oc` command instead of `kubectl`, then `k8sClient.oc*()` should get an improvement to support that, and then call `k8sClient.oc_get("network-attachment-definition", oc_cmd="oc")`.

task.py
```diff
@@ -381,7 +382,21 @@ def get_pod_ip(self) -> str:
         pod_ip = None
         try:
             if y:
-                pod_ip = y["status"]["podIP"]
+                if self.ts.connection.secondary_network_nad:
```
thom311, line 385: This will currently always be True, because `secondary_network_nad` is filled with the default `f"{namespace}/ocp-secondary"`.

If you need to differentiate whether the caller explicitly configured the setting in YAML, you must make `secondary_network_nad: Optional[str]`, so that a `None` represents unset (and each callers need to figure out what actual value to use).

thom311:
Could you please in the commit message elaborate how this works? Is there a Jira issue to link to? Could you give some history of this feature? How to use this? Does the cluster need a special configuration, etc.

-- I am not questioning the purpose or validity of the change. But please record the purpose and the workings somewhere (possibly in the commit message).

vrindle:
@thom311 no problem. I will add the commit message.

task.py
vrindle: Ack I will do that.

testConfig.py
vrindle: okay I can use k8sclient.

testConfig.py
vrindle: Okay I can try to do it the first way! Thanks :)

thom311:
@vrindle could you please rebase to latest `main`? There were some changes on `testConfig.py`, that will require your patch to change. I hope it's clear what to adjust. Sorry about the inconvenience.

vrindle:
@thom311 no problem. I am working on fixing up this PR right now. I will rebase on top of main.

testConfig.py
```diff
@@ -262,7 +268,12 @@ def parse(
 
             secondary_network_nad = common.structparse_pop_str(
                 *varg.for_key("secondary_network_nad"),
-                default=f"{namespace}/ocp-secondary",
+                default=None,
```
thom311, line 271: there are various users of `secondary_network_nad`. For example, `Task.get_secondary_ip()` would need to handle a `None` value. Only some of the users care to see whether the value was explicitly set.

Maybe:
```diff
diff --git i/testConfig.py w/testConfig.py
index 68338a9db8e5..45fcc67196aa 100644
--- i/testConfig.py
+++ w/testConfig.py
@@ -199,6 +199,7 @@ class ConfConnection(StructParseBaseNamed):
     plugins: tuple[ConfPlugin, ...]
     secondary_network_nad: Optional[str]
     resource_name: Optional[str]
+    namespace: str
 
     def serialize(self) -> dict[str, Any]:
         extra: dict[str, Any] = {}
@@ -312,8 +313,15 @@ class ConfConnection(StructParseBaseNamed):
             plugins=plugins,
             secondary_network_nad=secondary_network_nad,
             resource_name=resource_name,
+            namespace=namespace,
         )
 
+    @property
+    def secondary_network_nad_or_default(self) -> str:
+        if self.secondary_network_nad is None:
+            return f"{self.namespace}/ocp-secondary"
+        return self.secondary_network_nad
+
 
 @strict_dataclass
 @dataclass(frozen=True, kw_only=True)
```
and let some of the callers use `secondary_network_nad_or_default`.

task.py
```diff
@@ -275,16 +276,40 @@ def get_namespace(self) -> str:
...
+            return None
+
     def get_template_args(self) -> dict[str, str | list[str]]:
+        extra: dict[str, Any] = {}
+        if isinstance(self.ts.connection.secondary_network_nad, str):
+            common.dict_add_optional(
+                extra, "secondary_network_nad", self.ts.connection.secondary_network_nad
+            )
+        if isinstance(self.ts.connection.resource_name, str):
+            common.dict_add_optional(
```
thom311, line 300: The `if isinstance` check is not necessary.

The point of `common.dict_add_optional()` is to take an `Optional[T]` value, and add the entry to the dict if-and-only-if the value is not None.

task.py
```diff
@@ -275,16 +276,40 @@ def get_namespace(self) -> str:
...
-            "image_pull_policy": tftbase.get_tft_image_pull_policy(),
-            "command": ["/usr/bin/container-entry-point.sh"],
-            "args": [],
-            "index": f"{self.index}",
-            "node_name": self.node_name,
-            "secondary_network_nad": self.ts.connection.secondary_network_nad,
+            "name_space": self.get_namespace(),  # str
+            "test_image": tftbase.get_tft_test_image(),  # str
+            "image_pull_policy": tftbase.get_tft_image_pull_policy(),  # str
+            "command": ["/usr/bin/container-entry-point.sh"],  # str
```
thom311, line 308: what is the meaning of those `# str` comments?

task.py
```diff
@@ -381,7 +406,21 @@ def get_pod_ip(self) -> str:
...
+                if self.ts.connection.secondary_network_nad:
+                    network_status_str = y["metadata"]["annotations"][
+                        "k8s.v1.cni.cncf.io/network-status"
+                    ]
+                    network_status = json.loads(network_status_str)
+
+                    for network in network_status:
+                        if (
+                            network["name"]
+                            == f"default/{self.ts.connection.secondary_network_nad}"
```
thom311, line 418: previously, `secondary_network_nad` would already include the namespace. So it would default to `f"{namespace}/ocp-secondary" (where `namespace` defaults to `"default"`).

Here, the namespace gets duplicated, that seems wrong (unless you change the meaning of the attribute to not include a namespace).

task.py
```diff
@@ -275,16 +276,40 @@ def get_namespace(self) -> str:
...
+    def _fetch_default_resource_name(self) -> Union[str, None]:
+        k8sClient = K8sClient()
+        data = k8sClient.oc_get("network-attachment-definition")
+
+        if isinstance(data, dict):
+            resource_name = data["items"][0]["metadata"]["annotations"].get(
+                "k8s.v1.cni.cncf.io/resourceName"
+            )
+
+        if isinstance(resource_name, str):
```
thom311, line 288: Here there is some error checking that `resource_name` might not be a string. But there are many other conditions, in which `"data["items"][0]["metadata"]["annotations"].get(...)` can fail.

Maybe something like:
```diff
diff --git i/task.py w/task.py
index eb45eae9ca8c..ef42d8de326e 100644
--- i/task.py
+++ w/task.py
@@ -280,15 +280,17 @@ class Task(ABC):
         k8sClient = K8sClient()
         data = k8sClient.oc_get("network-attachment-definition")
 
-        if isinstance(data, dict):
-            resource_name = data["items"][0]["metadata"]["annotations"].get(
-                "k8s.v1.cni.cncf.io/resourceName"
-            )
-
-        if isinstance(resource_name, str):
-            return resource_name
-        else:
-            return None
+        resource_name = None
+        if data is not None:
+            try:
+                r = data["items"][0]["metadata"]["annotations"].get(
+                    "k8s.v1.cni.cncf.io/resourceName"
+                )
+                if isinstance(r, str) and r:
+                    resource_name = r
+            except Exception:
+                pass
+        return resource_name
 
     def get_template_args(self) -> dict[str, str | list[str]]:
         extra: dict[str, Any] = {}
```

task.py
```diff
@@ -275,7 +276,32 @@ def get_namespace(self) -> str:
...
+
+        if data is not None:
+            try:
+                r = data["items"][0]["metadata"]["annotations"].get(
+                    "k8s.v1.cni.cncf.io/resourceName"
+                )
+                if isinstance(r, str) and r:
+                    resource_name = r
+            except Exception:
+                pass
```
wizhaoredhat, line 291: Why are we passing on Exception?

vrindle:
**Note this code was also tested and worked properly.

task.py
wizhaoredhat: We can't use `self.namespace` here?

wizhaoredhat (commented):
Overall looks good, just a few minor questions and comments.

task.py
vrindle: really good point @wizhaoredhat! I am updating that right now.

task.py
vrindle: We pass here because we don't need to anything to resource_name if there is an exception (resource name will be None and that is what we will return).

wizhaoredhat: Wait shouldn't this function use "secondary_network_nad" ?

task.py
thom311: > The secondary_network_nad did not already include the namespace... earlier we were setting default value to include the namespace but secondary_network_nad if set by user will be whatever value the user inputs.

The default value included the namespace by having the format "NAMESPACE/NAME".
Of course, a user who configures the setting in YAML, can set it to any string. But that will only work if they follow the same format as intended for this property (and the same format that the default value does).

You can add some code, to change to specify this without namespace (or optionally with/without namespace). But there needs to be clear what this property is supposed to be, and unless you change that, it's NAMESPACE/NAME.

testConfig.py
```diff
@@ -262,7 +276,17 @@ def parse(
...
+            )
+
+            resource_name = common.structparse_pop_str(
+                *varg.for_key("resource_name"),
+                default=None,
+            )
+
+            namespace = common.structparse_pop_str(
+                *varg.for_key("namespace"),
+                default="default",
```
thom311, line 289: I think you should delete this `structparse_pop_str("namespace")` and delete serializing the namespace in `serialize()`.

Note that `namespace` is already an argument passed in by the caller. While this value comes from YAML, it is defined at a parent/outer/caller level and not here.

These `Conf*` classes mainly aim to represent the content of the YAML. Here, `ConfConnection.namespace` is not actually present in the YAML (at this layer). That is slightly ugly, but OK. Note if you parse the entire YAML, then the value is still fully determined, since the value comes from the YAML (just from another place).

You can come up with schemes how we could also add a namespace to YAML at this place, in particular, you would then here need to set `default=namespace`. But overall, it seems unnecessary. `ConfConnection.namespace` is just a bit odd, and not in YAML (at this layer). It's fine.

task.py
vrindle: @thom311 we hardcode the value... can I just make it so that it is "ocp-secondary" instead of "default/ocp-secondary"? I have already added a way for it use `self.namespace` is that function.

task.py
```diff
@@ -275,7 +276,32 @@ def get_namespace(self) -> str:
     def get_duration(self) -> int:
         return self.ts.cfg_descr.get_tft().duration
 
+    def _fetch_default_resource_name(self) -> Union[str, None]:
```
thom311, line 279: could we do:
```diff
-    def _fetch_default_resource_name(self) -> Union[str, None]:
-        data = self.client.oc_get("network-attachment-definition")
+    @staticmethod
+    @functools.cache
+    def _fetch_default_resource_name(client: K8sClient) -> Optional[str]:
+        data = client.oc_get("network-attachment-definition")
         resource_name = None
```

The purpose is to determine the value only once per run, instead of multiple times (and getting potentially different results). Also, each `client.oc_get()` will log data, which is good, but seeing it redundantly is confusing.

config.yaml
```diff
@@ -22,7 +22,9 @@ tft:
           - name: measure_cpu
           - name: measure_power
           - name: validate_offload
-        # Secondary network is required for tests 27, 28 and 29
+        # (Optional) Secondary network is required for tests 27, 28 and 29. Defaults to None if not set.
```
thom311, line 25: this `Defaults to None if not set` is not right.

The callers of `secondary_network_nad_or_default` have a default as `f"{namespace}/ocp-secondary"` (where `namespace` itself can be configured in YAML but defaults to `"default"`).

The callers that directly use ` ConfConnection.secondary_network_nad` will do something specific about `None` (`pod_ip = y["status"]["podIP"]`), but saying that the default is "None" is not meaningful. For this case, the description should say something like: `If configured, "secondary_network_nad" is also used to detect the pod IP"`.

But can we take one step back? Why `get_pod_ip()` even does this? A value in YAML having a default value, is for convenience of the user so that they can type fewer settings in their configuration. But the behavior should be the same, whether they explicitly type a value or get the default. `get_pod_ip()` should not care about whether the value is explicitly configured or just happens to be the default. This overloads the meaning of whether the value is configured, in confusing ways. If this should be configurable, then add a boolean key "get_pod_ips_from_secondary_nad" and check that in `get_pod_ip()`. Or quite possibly, the test should figure this out by itself without making it configurable to the user.

thom311: same in `Task.get_template_args()` which does
```
task.py-        common.dict_add_optional(
task.py-            extra, "secondary_network_nad", self.ts.connection.secondary_network_nad
task.py-        )
```
instead of `secondary_network_nad_or_default`.

This changes behavior when rendering `manifests/allow-egress-mnp.yaml.j2`.

task.py
thom311: > we hardcode the value... can I just make it so that it is "ocp-secondary" instead of "default/ocp-secondary"?

Sure it can change, but then the various current callers need to be updated to correctly handle that. And it must be consistently used. And maybe add a check that `if "/" in secondary_network_nad: raise ValueError(f"\"{yamlpath}.secondary_network_nad: cannot contain a namespace")`

But see also https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/141#discussion_r1762472795 . I just realized, that IMO we shouldn't try to attribute additional meaning to whether the value is explicitly configured or at the default.

config.yaml
thom311: ok, we discussed this. So "secondary_network_nad" will just behave differently, depending on the test. The README should explain how it is, roughly:

- for tests 27-29, they always need such a NAD. Thus, if left unconfigured, they use a default of `f"{namespace}/ocp-secondary"`.
- in other cases, if `secondary_network_nad` property is explicitly configured in YAML, some (all? which?) tests will use that to use the secondary interface.

Note that with `tft[*].test_cases = "*"` we can run multiple test cases for one test configuration. How it makes sense to configure/omit `secondary_network_nad` value while running test cases that treat the value differently, is up to the user. Maybe it will be common to have configurations with `test_cases = 27-29` or `test_cases = 0-26,30-100`.

task.py
vrindle: working on adding that!

vrindle:
** Note this has been tested and runs properly.

wizhaoredhat:
@thom311 PTAL.

task.py
```diff
@@ -381,7 +420,23 @@ def get_pod_ip(self) -> str:
...
+                if self.ts.connection.secondary_network_nad:
+                    network_status_str = y["metadata"]["annotations"][
+                        "k8s.v1.cni.cncf.io/network-status"
+                    ]
+                    network_status = json.loads(network_status_str)
+
+                    namespace = self.get_namespace()
+
+                    for network in network_status:
+                        if (
```
wizhaoredhat, line 432: NIT: This looks kind of bad splitting it like this...

task.py
```diff
@@ -275,7 +277,43 @@ def get_namespace(self) -> str:
...
+                    resource_name = r
+                    logger.info(f"autodetected resource_name as {resource_name}")
+            except Exception:
+                pass
+        return resource_name
+
     def get_template_args(self) -> dict[str, str | list[str]]:
+        extra: dict[str, Any] = {}
+        common.dict_add_optional(
+            extra, "secondary_network_nad", self.ts.connection.secondary_network_nad
```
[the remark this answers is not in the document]
vrindle: it does work with no namespace

thom311 (approved):

wizhaoredhat:
LGTM

PR #143, opened 2024-09-12
[th/firewall-doc] document firewall methods and minor improvements

thom311:
Commits:

- host: let host.Result evaluate in a boolean context
- firewall: add docs for nft_data_masquerade_up() and nft_call()
- firewall: return host.Result from firewall.nft_call() and take shell script

Add doc since the usage may not be obvious. Also, add a unit test, which might help to see the usage.


wizhaoredhat:
LGTM! Thanks!

PR #144, opened 2024-09-12
[th/containerfile-workdir] Containerfile: set default workdir to "/opt/ocp-tft" and install nftables

thom311:
these two things seem potentially useful to have in our container.

- install `nftables`. It adds 1.4MB, but IMO size of the container is not a major goal, but being useful is.
- set `WORDIR /opt/ocp-tft`. This way, we can directly `python -c "import ...` our Python code. It's more useful than `/` as working directory.


wizhaoredhat:
LGTM

PR #145, opened 2024-09-16
[th/iter-get-first] common: support requesting exactly one element in common.iter_get_first()

thom311:
- add special case "unique=True, force_unique=True" to reject empty lists.

- rework to use next(iter()) instead of a for loop.


wizhaoredhat:
I think we need some commentary to explain the usage for:
    unique: bool = False,
    force_unique: bool = False,
    single: bool = False,
  It is not obvious why these parameters are optionally needed.

wizhaoredhat:
LGTM!

PR #146, opened 2024-09-18
[th/ns-for-secondary-nad] task: improve "secondary_network_nad" for being with/without namespace

thom311:
Note that "default-network" defaults to "default/default", it thus commonly includes the namespace. Interestingly, the namespace is not the one configured for the test run, but that seems intentional since CDA creates a NAD "default/default". Anyway. Most likely setting "default-network" works both with or without a namespace. If the namespace is omitted, the NAD should be relative to the test's namespace.

For "secondary_network_nad", this should be handled the same. The user shall be able to configure any namespace. They may also omit the namespace, in which case the NAD is in the test's namespace. For tests 27-29, the default is "ocp-secondary" (which is again in the test's namespace).

Adjusts various places so that "secondary_network_nad" works with those assumptions.

CC: @vrindle

### DIFF
--- a/task.py
+++ b/task.py
@@ -282,9 +282,13 @@ class Task(ABC):
         client: K8sClient, namespace: str, secondary_network_nad: Optional[str]
     ) -> Optional[str]:
         if secondary_network_nad is not None:
+            if "/" in secondary_network_nad:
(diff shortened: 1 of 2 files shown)


wizhaoredhat:
LGTM

PR #147, opened 2024-09-19
[th/host-log-with-output] host: print full result on error even with log_lineouput enabled

thom311:
With host.run(..., log_lineoutput=True), we log the output line by line as it happens. Usually, if we already do this, we don't want to repeat logging the output in the summary line. Especially, because we expect the output to be of a long running process that shows it's progress. When a caller sets log_lineoutput, we expect the output to be long, and usually not as interesting (as for example host.local.run("ip -j link") to gather information we use).

However, if the logging level of the result is higher than the level of the log lines, then the user may not have the individual lines printed in the output. More importantly, the logging level of the result is likely higher because there was an error (e.g. with "die_on_error" or "log_level_fail"). In those cases, we want to make sure that we can see the output of the command that just failed.

Fix this by including the full process output in the summary line in some cases, even if log_lineoutput is enabled.

  $ python -c 'if 1:
      import ktoolbox.host as h

      h.local.run("cat pytest.ini; exit 1", log_lineoutput=True)
      h.local.run("cat pytest.ini; exit 1", log_lineoutput=True, die_on_error=True)
  '
  2024-09-19 09:46:29.430 DEBUG   [th:140011636640640]: cmd[1;localhost]: call 'cat pytest.ini; exit 1'
  2024-09-19 09:46:29.432 DEBUG   [th:140011636640640]: cmd[1;localhost]: stdout[0] b'[pytest]\n'
  2024-09-19 09:46:29.432 DEBUG   [th:140011636640640]: cmd[1;localhost]: stdout[1] b'python_classes = !Test*\n'
  2024-09-19 09:46:29.432 DEBUG   [th:140011636640640]: cmd[1;localhost]: └──> 'cat pytest.ini; exit 1': failed (exit 1)
  2024-09-19 09:46:29.432 DEBUG   [th:140011636640640]: cmd[2;localhost]: call 'cat pytest.ini; exit 1'
  2024-09-19 09:46:29.433 DEBUG   [th:140011636640640]: cmd[2;localhost]: stdout[0] b'[pytest]\n'
  2024-09-19 09:46:29.433 DEBUG   [th:140011636640640]: cmd[2;localhost]: stdout[1] b'python_classes = !Test*\n'
  2024-09-19 09:46:29.433 ERROR   [th:140011636640640]: cmd[2;localhost]: └──> 'cat pytest.ini; exit 1': [FATAL] failed (exit 1); out='[pytest]\npython_classes = !Test*\n'
  2024-09-19 09:46:29.434 ERROR   [th:140011636640640]: FATAL ERROR. BACKTRACE:
    File "<string>", line 5, in <module>
    File "/data/src/ocp-traffic-flow-tests/ktoolbox/host.py", line 505, in run
      f"FATAL ERROR. BACKTRACE:\n{''.join(traceback.format_stack())}"


wizhaoredhat:
LGTM

PR #148, opened 2024-09-26
[th/delete-ignore-noexist] prevent error message in logfile when deleting non-existing `multi-networkpolicies`

thom311:
During cleanup, we always call `kubectl --kubeconfig /root/kubeconfig.nicmodecluster -n default delete multi-networkpolicies -l tft-tests`. But for most test cases we won't have such a resource, so the command fails.

Since `k8sClient.oc()` by default has `may_fail=False`, this results in an error in the log:

```
2024-09-16 20:19:03.270 INFO    [th:140347998742336]: Cleaning pods, services and multi-networkpolicies with label tft-tests in namespace default
2024-09-16 20:19:04.467 ERROR   [th:140347998742336]: cmd[8435;localhost]: └──> 'kubectl --kubeconfig /root/kubeconfig.nicmodecluster -n default delete multi-networkpolicies -l tft-tests': failed (exit 1); err='error: the server doesn\'t have a resource type "multi-networkpolicies"\n'
```

It's important that the log file is not spammed with error messages for things that are no errors. Do something about that.

See: https://net-hw-ci-jenkins-csb-nst.apps.ocp-c1.prod.psi.redhat.com/view/Custom%20OVNK8%20Testing%20(NIC%20Mode)/job/99_Lab227_New_OVN_Traffic_Flow/189/consoleFull

### DIFF
--- a/ktoolbox/host.py
+++ b/ktoolbox/host.py
@@ -183,6 +183,7 @@ class BinResult(BaseResult[bytes]):
             self.out.decode(errors=errors),
             self.err.decode(errors=errors),
             self.returncode,
+            forced_success=self.forced_success,
         )
 
     def dup_with_forced_success(self, forced_success: bool) -> "BinResult":
--- a/ktoolbox/host.py
+++ b/ktoolbox/host.py
@@ -1,6 +1,7 @@
 import dataclasses
 import logging
 import os
+import re
 import select
 import shlex
 import subprocess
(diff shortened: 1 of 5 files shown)


wizhaoredhat:
This is definitely one way of doing this. 
I was going to say shouldn't we have some condition based on the config whether we try to run this cleanup action.
But then I do like this because it can potentially cleanup something from a previous run that was unexpectedly aborted.

So LGTM from my point of view.

thom311:
> This is definitely one way of doing this. I was going to say shouldn't we have some condition based on the config whether we try to run this cleanup action. But then I do like this because it can potentially cleanup something from a previous run that was unexpectedly aborted.

I think cleanup and prepare steps should be idempotent and we just run them eagerly. They should do nothing, if there is nothing to cleanup/prepare. The benefit is that we can cleanup a crash from an earlier run.

PR #149, opened 2024-09-26
[th/show-failure-reason] evalutor: print the reason for the test failure

thom311:
It can be hard to understand why the failure happened. Track the reason and also show it in "print_results.py"

The test result already contains a "success" field, and a reason "msg" if the test failed. But if the failure happens during evaluation and the bitrate threshold, it's hard to understand.

For example, "print_results.sh" prints now:
```
  There are 4 failing flows. Details:
  "Test ID: HOST_TO_CLUSTER_IP_TO_POD_DIFF_NODE, Test Type: IPERF_TCP, Reverse: false, TX Bitrate: 11.649 Gbps, RX Bitrate: 11.633 Gbps, Run succeeded but Bitrate(tx=11.649, rx=11.633) is below threshold 18.0"
  ...
```


wizhaoredhat:
Very good! LGTM

PR #150, opened 2024-09-30
[th/rework-logger] logger: rework logging and drop ktoolbox.logger

thom311:
Python's logging module provides hierarchical loggers. Use them as intended, that is, different modules use different loggers, which can be configured individually (hierarchically). This is instead of all modules sharing the `ktoolbox.logger` instance.

Users of logging, just import `logging` and get their logger via `logging.getLogger(__name__)`. Note that for the modules of the main application, we don't have a package name. That means, it would be cumbersome to configure them all at once. We work around that by instead using `logging.getLogger("tft." + __name__)`, and configure the "tft" logger(s).

The logging now only gets explicitly configured by the main modules. They use helper methods `common.log_config_loggers()` for that. That one also honors the environment variable `KTOOLBOX_LOG_LEVEL` to determine the default behavior if the logging level is unspecified on the command line.

With `KTOOLBOX_ALL_LOGGERS` environment variable set to True, `common.log_config_loggers()` will configure the root logger instead. This means that also other python modules may have their logging enabled.


wizhaoredhat:
LGTM

PR #151, opened 2024-10-02
[th/no-deprecated-type-aliases] ktoolbox: don't use deprecated typing.{List,Dict,Tuple,Type}

thom311:
Since Python 3.9, the builtin types {list,dict,tuple,type} support generic type subscription. Use those instead of the deprecated variants.

https://docs.python.org/3/library/typing.html#aliases-to-built-in-types


wizhaoredhat:
LGTM

PR #152, opened 2024-10-03
[th/fix-log-config] common: improve log_config_logger() for multiple calls

thom311:
- previously, common.log_config_loggers() would each time add a new handler. That meant, if you call configure logger more than once, multiple handlers are added and multiple lines are printed.

  This is not uncommon, because we might want initially setup one logging level, then read further configuration, and reconfigure the loggers with the new level.

  Instead, the handler that we add now has a particular internal type common._LogHandler. If such a handler is already registered, subsequent calls will not create a new handler but simply reconfigure the existing one.

- rename common.log_config_loggers() to common.log_config_logger(). While the function takes a variadic list of loggers as argument to configure any number of loggers, this seems the better name. This means to merge common.log_config_logger() and common.log_config_loggers(). There is now only one variant that supports both use cases.

### DIFF
--- a/evaluator.py
+++ b/evaluator.py
@@ -176,7 +176,7 @@ def parse_args() -> argparse.Namespace:
 
     args = parser.parse_args()
 
-    common.log_config_loggers(args.verbose, "tft", "ktoolbox")
+    common.log_config_logger(args.verbose, "tft", "ktoolbox")
 
     if not Path(args.config).exists():
         logger.error(f"No config file found at {args.config}, exiting")
--- a/ktoolbox/common.py
+++ b/ktoolbox/common.py
@@ -8,6 +8,7 @@ import os
 import re
 import time
 import typing
+import threading
 
 from collections.abc import Iterable
 from collections.abc import Mapping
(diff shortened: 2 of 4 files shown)


wizhaoredhat:
LGTM

wizhaoredhat:
But tests are failing?

wizhaoredhat:
LGTM - That type checking thing is quite nasty looking but it is the solution that python developers are saying to use.

thom311:
> But tests are failing?

Yes, I didn't notice before. It failed with Python3.9, which is the oldest supported version of the `ktoolbox` module. Luckily this problem was uncovered by tests here.

Should be fixed now, it was due what is explained in the [docs](https://mypy.readthedocs.io/en/stable/runtime_troubles.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime) for `logging.StreamHandler` vs `logging.StreamHandler[TextIO]`.

wizhaoredhat:
LGTM. Locking is definitely needed, nice catch.

PR #153, opened 2024-10-03
common: add ExtendedLogger() class

thom311:
The benefit of an ExtendedLogger instance is that it provides additional convenience methods (currently only `error_and_exit()`).

All other Logger methods are delegated to the wrapped logger instance. As the method `error_and_exit()` is stateless, there is no difference between using the extended logger or the wrapped logger.

Hence, with

    ext_logger = common.ExtendedLogger("foo")
    real_logger = logging.getLogger("foo")

calling a method like `debug()` will yield the same result. That is a promise.

The benefit here is that the extended logger has convenience methods on top. While the user does not get access to them via `logging.getLogger()`, they can create (and cache) an `ExtendedLogger` instance for that. The linter will nicely understand when you may call `logger.error_and_exit()`.

### DIFF
--- a/ktoolbox/common.py
+++ b/ktoolbox/common.py
@@ -6,9 +6,10 @@ import json
 import logging
 import os
 import re
+import sys
+import threading
 import time
 import typing
-import threading
 
 from collections.abc import Iterable
 from collections.abc import Mapping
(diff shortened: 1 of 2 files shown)


wizhaoredhat:
LGTM

PR #156, opened 2024-10-17
Use Github Container Repository

flavio-fernandes:
In anticipation to this repo moving... ;)

### DIFF
--- a/.github/workflows/build-and-push.yml
+++ b/.github/workflows/build-and-push.yml
@@ -16,16 +16,14 @@ jobs:
     - name: Set up Docker Buildx
       run: |
         docker buildx create --use
-    - name: Authenticate
-      uses: docker/login-action@v1
-      with:
-        registry: quay.io
-        username: ${{ secrets.QUAY_USERNAME }}
-        password: ${{ secrets.QUAY_PASSWORD }}
-    - name: Build and Push Multiarch image
+    - name: Log in to GitHub Container Registry
+      run: |
+        echo "${{ secrets.GITHUB_TOKEN }}" | docker login ghcr.io -u ${{ github.actor }} --password-stdin
+    - name: Build and Push Multiarch Image to ghcr.io
       run: |
         docker buildx build \
           --file ./Containerfile \
-          --platform linux/arm64,linux/amd64\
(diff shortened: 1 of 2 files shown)


wizhaoredhat:
@thom311  / @SalDaniele  PTAL

SalDaniele:
LGTM Thanks @flavio-fernandes!

thom311:
Could you please elaborate on the background of this patch? Why do we expect the repo to be moving?

flavio-fernandes:
> Could you please elaborate on the background of this patch? Why do we expect the repo to be moving?

Hi @thom311,
I might be mistaken here, so please feel free to correct me, but I was under the impression that traffic-flow-tests would eventually become part of the ovn-kubernetes umbrella, as the project transitions into its own organization [ [1](https://github.com/cncf/sandbox/issues/111) ]. While this change is still a bit down the road, my current goal is to integrate this fantastic suite as a lane in the ovn-kubernetes pull requests. This way, we can ensure no regressions occur concerning the expected traffic flows that should always function properly [ [2](https://github.com/ovn-org/ovn-kubernetes/issues/4756) ].

[1]: https://github.com/cncf/sandbox/issues/111
[2]: https://github.com/ovn-org/ovn-kubernetes/issues/4756

wizhaoredhat:
@flavio-fernandes Do I need to install "secrets.GITHUB_TOKEN" somehow?

SalDaniele:
For my fork I had to add a new Personal access token to my Github:
- Settings
- Developer Settings
- Personal Access Tokens (opted for classic)

Copy the token and add it as a secret to the repo as we did with quay credentials in the past (Will push a follow-up PR because it seems Github does not like secrets starting with "GITHUB")

SalDaniele:
Actually disregard, I think this secret exists by default.

https://github.com/wizhaoredhat/ocp-traffic-flow-tests/pull/158 should do the trick

flavio-fernandes:
> Actually disregard, I think this secret exists by default.
> 
> #158 should do the trick

TY both!

flavio-fernandes:
I can now `docker pull ghcr.io/wizhaoredhat/ocp-traffic-flow-tests:pr_test`   woot. ;)

PR #157, opened 2024-10-18
[th/rework-register-modules] rework how TestTypeHandler and pluginbase.Plugin modules register

thom311:
I think this approach is simpler. That is, it's harder to get wrong and it's fewer lines of code.

That's interesting, when we add more test-types or plugins.

### DIFF
--- a/testTypeIperf.py
+++ b/testTypeIperf.py
@@ -64,8 +64,8 @@ class TestTypeHandlerIperf(TestTypeHandler):
         return False
 
 
-test_type_handler_iperf_tcp = TestTypeHandlerIperf(TestType.IPERF_TCP)
-test_type_handler_iperf_udp = TestTypeHandlerIperf(TestType.IPERF_UDP)
+TestTypeHandler.register_test_type(TestTypeHandlerIperf(TestType.IPERF_TCP))
+TestTypeHandler.register_test_type(TestTypeHandlerIperf(TestType.IPERF_UDP))
 
 
 class IperfServer(task.ServerTask):
--- a/pluginMeasureCpu.py
+++ b/pluginMeasureCpu.py
@@ -37,7 +37,7 @@ class PluginMeasureCpu(pluginbase.Plugin):
         ]
 
 
-plugin = PluginMeasureCpu()
+plugin = pluginbase.register_plugin(PluginMeasureCpu())
 
 
 class TaskMeasureCPU(PluginTask):
--- a/pluginMeasurePower.py
+++ b/pluginMeasurePower.py
@@ -40,7 +40,7 @@ class PluginMeasurePower(pluginbase.Plugin):
         ]
 
 
-plugin = PluginMeasurePower()
+plugin = pluginbase.register_plugin(PluginMeasurePower())
 
 
 def _extract(ipmitool_output: str) -> Optional[int]:
--- a/pluginValidateOffload.py
+++ b/pluginValidateOffload.py
@@ -136,7 +136,7 @@ class PluginValidateOffload(pluginbase.Plugin):
         ]
 
 
-plugin = PluginValidateOffload()
+plugin = pluginbase.register_plugin(PluginValidateOffload())
 
 
 class TaskValidateOffload(PluginTask):
--- a/pluginbase.py
+++ b/pluginbase.py
@@ -1,4 +1,5 @@
 import logging
+import threading
 import typing
 
 from abc import ABC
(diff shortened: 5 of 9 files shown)


tftbase.py
```diff
@@ -21,6 +22,7 @@
 
 logger = logging.getLogger("tft." + __name__)
 
+tft_lock = threading.Lock()
```
wizhaoredhat, line 25: Why do we need this lock? What is being thread here during config?

thom311: hm...

well, with this PR, registration happens when the modules are imported (and call `pluginbase.register_plugin()` and `TestTypeHandler.register_test_type()`). There is also only one place when we import the modules, and that is during `pluginbase._get_registry()` and `TestTypeHandler.get()` (although, the code really should also work if you import the modules manually and explicitly).

ocp-tft is multi threaded. Still, I would very much suspect that above getters are called before other threads start, so most likely there should not be a problem. At least "today". This is the important part, because it adds hidden requirements like "ensure that the modules are imported on a thread before other threads are started". We have a mult-threaded application, and this plugin/module registration necessarily uses global variables. I think there should be a lock around accessing them just as a matter of principle.

---

Usually, when we want to lock one thing (e.g. access to `pluginbase._registry`), then we create a dedicated lock instance. In this case, I instead reused the same lock `tftbase.tft_lock` for different purposes. I think it's fine because:
- whenever we hold a lock we always must ensure that we don't call into unknown code (e.g. use provided callbacks), so that we don't take the same lock again (`threading.lock` is not reentrant) and don't have multiple locks that are taken in undefined order (danger of deadlock). In our use here, it's easy to see that we don't call to code that could lead to reentrantly taking a lock or to deadlock.
- using the same lock might decrease parallelism, but in practice, this doesn't matter here.

---

Anyway. I drop `tftbase.tft_lock` and add instead two specific locks per usage. That makes it clearer for when each lock is used.

testTypeIperf.py
```diff
@@ -64,8 +64,8 @@ def can_run_reverse(self) -> bool:
         return False
 
 
-test_type_handler_iperf_tcp = TestTypeHandlerIperf(TestType.IPERF_TCP)
-test_type_handler_iperf_udp = TestTypeHandlerIperf(TestType.IPERF_UDP)
+TestTypeHandler.register_test_type(TestTypeHandlerIperf(TestType.IPERF_TCP))
```
wizhaoredhat, line 67: If we do "register_test_type" as part of the main function prior to creating threads. Then we don't need the lock right?

I don't think the register functions need to be called per file. We can consolidate it all in one place as part of the start up of the scripts.

(This comment applies to plugins as well)

What is the purpose of registering? Are we expecting users to remove plugins?

thom311: > What is the purpose of registering? Are we expecting users to remove plugins?

We have a `get()` method, that can lookup plugins/test-type-handlers by name/ID. It could be implemented in various ways. For example, previously `TestTypeHandler.get()` had an `if-elif-else` block and named every ID. This now changed to use a dictionary. I think that is slightly nicer (see below).

> If we do "register_test_type" as part of the main function prior to creating threads. Then we don't need the lock right?

We could require that the user calls a setup/initialize method first before being allowed to call `get()`. This pushes an additional requirement to the user (note that `main.py`, `evaluator.py` and unit tests would all need to follow this requirement). I think it is nicer that any caller can just call `get()` at any time (and on any thread), and it will initialize itself in a thread-safe way. One thing less a user can do wrong.

---

This PR moves some code around and changes who does what where. It's not a major change, but I think this is nicer because:

- it's fewer lines of code, which is often an indication that it's better
- if we add a new plugin/test-type, then we require only one single line outside of the plugin/test-type module (the `import newModule`). Additionally, it needs the `register()` call, but that's inside the new module itself.
- we create new plugin/test-type via copy+paste. It's pretty obvious how you must call the `register()` method (just copy and adjust). Also, there are unit tests, so it would be pretty hard to create a new plugin/test-type and make a mistake in how it registers itself.

No major difference, but slightly nicer IMO.

thom311: I want that there is less work when a new plugin/test-types is added.

With this PR, a plugin/test-type requires one `import newModule` and one `register()` call to be hooked up. We could drop the `import newModule` if we would search the file system for `plugin*.py`/`testType*.py` files. But that seems worse, because we should be explicit which modules are loaded. So two lines is minimal.

wizhaoredhat:
ok, it is not too bad. I think having the singular global is ok here. The reason you have given makes sense, it does definitely reduce the amount of code for sure.

wizhaoredhat:
LGTM

PR #160, opened 2024-11-15
[th/config-relative-paths] better handle relative paths

thom311:
We can have relative paths both inside the YAML configuration and as command line arguments.

`TestConfig` should make them absolute.

First, let each `Conf*` get a reference to the parent/owner object. This way, from a nested object of the configuration, we can get up to the top-level entry. The use is in this case that the top level entry tracks the directory name, for which we want to interpret relative paths.

Then, kubeconfig can either be in the YAML configuration or in the command line argument (well, technically, the command line argument is not yet implemented). But the point is, there are potentially two different directories which is the base for relative paths, depending on which path we talk about. A kubeconfig path in the YAML config should be relative to the YAML config file. A kubeconfig path in the command line should be relative to the current working directory. For that, add two properties `TestConfig.configdir` and `TestConfig.cwddir`.

The `logs` property is in the YAML configuration, bit it's relative to the current working directory. Add a `logs_absdir` property, that is resolved to be absolute.


wizhaoredhat:
I don't really understand what benefits we get from this.

thom311:
>I don't really understand what benefits we get from this.

Primarily, it's about the paths. The `logs` option is documented to be relative to the "working directory". The `kubeconfig` option should be relative to the YAML config file. With this branch, `TestConfig` tracks the base directories `cwddir` and `configpath`. Those are the directories for resolving the relative paths. Also, from deeper nested entries, we can now access the parent node, so that

```python
    @property
    def logs_abspath(self) -> pathlib.Path:
        return common.path_norm(
            self.logs,
            cwd=self.config.test_config.cwddir,
        )
```

works.

PR #161, opened 2024-11-19
[th/print-results-improvements] rework how print_results.py shows the result

thom311:
an attempt to make the results easier to read

- for `validate_offload` plugin, give a better error reason that shows the passed traffic and threshold.
- in `PluginResult` (which is also persisted to JSON), show the `msg` on failure and the full `plugin_result` details.
- rework how `print_result.sh` shows the output. In particular, print the plugin result. Also, if a plugin indicates failure, the entire run is marked as failed.

Related to https://issues.redhat.com/browse/NHE-1153 , but does not fully fix it.

### DIFF
--- a/pluginValidateOffload.py
+++ b/pluginValidateOffload.py
@@ -112,7 +112,7 @@ def check_no_traffic_on_vf_rep(
             return f"missing ethtool output for {direction}"
         return None
     if end - start >= VF_REP_TRAFFIC_THRESHOLD:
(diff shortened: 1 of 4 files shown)


thom311:
@SalDaniele how do you like this? (especially the last commit, see example in the commit message).

SalDaniele:
Nice! LGTM

wizhaoredhat:
LGTM

PR #162, opened 2024-11-20
[th/task-ignore-cancel-cmd-error] task: ignore errors from failed commands in task operation

thom311:
For running the command, we already ignored the exit code of the command. That is, because usually the command will be stopped by killing the process, and it's often that the exit code is non-zero.

For the stop/cancel command, now also ignore the exit code of the command. Seems there are cases where `killall iperf3` can fail. Note that it can be that the other side already terminated, which causes the iperf3 process to quit (--one-off) and killall to fail. Ignore that condition too.

### DIFF
--- a/task.py
+++ b/task.py
@@ -822,29 +822,32 @@ class ServerTask(Task, ABC):
 
         logger.info(f"Running {cmd}")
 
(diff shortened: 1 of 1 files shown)


SalDaniele:
LGTM

wizhaoredhat:
LGTM

PR #163, opened 2024-11-22
[th/structparse-context] use new StructParse API and helper objects

thom311:
There are two new types in `ktoolbox.common`: `StructParseParseContext` and
`StructParsePopContext`. They bundle `(yamlidx,yamlpath,arg)` and
`(vdict,yamlpath,key)`, respectively.
    
Those tuples of data are used repeatedly at various places,
`StructParseParseContext` is the argument to the `parse()` functions, and
`StructParsePopContext` is the argument to the `structparse_pop*()`
functions.
    
The API was updated with this new structures. Adjust.


SalDaniele:
It seems like bundling these common args puts the burden on the person writing the code to now understand this helper class and how to invoke it, versus the burden of having to pass these 3 args around constantly. I'm not sure if it is a worth while trade of in terms of readability if you are not familiar with ktools.

That said it does enforce the API nicely. So in short, fine with the change, LGTM

thom311:
> That said it does enforce the API nicely. So in short, fine with the change, LGTM

I think that's the main point.

Previously, I let them separate "plain" arguments, like the string we pass on. The idea was that you could call those function "directly" with the individual parameters at hand.

In practice, it makes the most sense to use this API in exactly the way it's used. Note that in the patch, there are almost no places where we explicitly need to create those bundle-objects. Instead, this adds some type safety and it also makes it clearer how to correctly use the API.

PR #164, opened 2024-11-26
[th/output-base-option] add `--output-base` option to pre-determine the name of the result files

thom311:
urrently, we have the "logs" configuration in the YAML config.
This defaults to "$PWD/ft-logs". The fact is that our YAML
config can contain more than one test configs ("tft-idx"). We
thus might write more than one result files and hence we allow
configuring the directory only.

But the chosen filenames contain timestamps. That is cumbersome as the
file name is no predictable from the user. The user will have to search
the output directory for files that look suitable.

Instead, add a command line option "--output-base" to "./main.py".
The user could use this like:

    OUTPUT_BASE="$PWD/ft-logs/result-"
    ./main config.yaml -o "$OUTPUT_BASE"

The benefit is that afterwards you know you can find the result files
as "${OUTPUT_BASE}*.json", and you know which tft-idx corresponds to
which file.

Note that if you provide a relative path (or just a basename without a
path separator), then the filename is relative to the current working
directory. We don't fallback to the "logs" path from the config as it
seems more expected that a command line argument is relative to the
current directory.  On the other hand, you can also specify a directory
name only (by appending a trailing '/'). That has the same effect as
setting the base for the filename to "result-".


SalDaniele:
Nice to have, LGTM

PR #167, opened 2024-11-27
[th/unify-output-formats] unify output formats for result and evaluated result

thom311:
Previously, we have `main.py`, which runs the test and writes a result file in
Format A (`TftResult`). Then, we have `evaluator.py`, which evaluates the
Format A file and marks runs as failed based on the threshold from
"eval-config.yaml". This second step results in a Format B
(`TestResultCollection`). Note that by passing an "eval-config.yaml"
parameter to `main.py`, both steps could be run in one, but that is only
a convenience feature and not fundamentally relevant. Finally, there is
`print_results.py` which can pretty print Format B.

That means, to pretty print the result, we always need to evaluate to
generate Format B. Even if we maybe don't have anything to evaluate because
we don't want to check for thresholds.

Also, the Format B is a restructuring of Format A. But not in a very
useful way, because `print_results.py` wants to show the output grouped
as it was in Format A, so it needs to undo the grouping of Format B (via
GroupedResult).

We don't need two formats. Instead, let `main.py` generate our Format A
(TftResult). At this point, we now always do also the evaluation (even
if the eval config might be empty). Later, we can run another
`evaluator.py` step so that the output gets again annotated/augmented
with the EvalResult. There really is only one format, and that is what
`main.py` produces. It's also what `evaluator.py` takes as input and
produces as output. Also, `print_results.py` operates on the one format.

This significantly simplifies the code. More importantly, it unifies
behaviors and code paths. And tools like `print_results.py` operate
on the result of `main.py` directly. And you can feed the output of
`evaluator.py` into `evaluator.py` again.


wizhaoredhat:
Thank you! This is a much needed change! LGTM

PR #168, opened 2024-11-27
[th/rename-main-to-tft] tft: rename "main.py" to "tft.py"

thom311:
We have multple python scripts that are directly executable (i.e. they have a `main()` function).

When running the test, it's not obvious why this program is called `main.py`. Rename it to `tft.py`. You should now run `tft.py`.

Add a symlink `main.py` that points to the new script. This is for backward compatibility, so that the various callers don't need to update their code (yet).


wizhaoredhat:
Strange that git doesn't think it was a file move. But LGTM

thom311:
> Strange that git doesn't think it was a file move. But LGTM

That's because the same commit moves the file and adds the file as symlink. In that case, git's heuristic doesn't detect merges.

That is indeed rather unfortunate, it makes the history a bit ugly. Note that `git log --follow tft.py` still works to walk the history back via the renamed file.

PR #169, opened 2024-11-27
[th/mypy-config] add pyproject.toml and move mypy.ini configurations there

thom311:
- pyproject: add "pyproject.toml" and move "mypy.ini" there
- github: don't pass addtional arguments to mypy in github workflows


wizhaoredhat:
LGTM

PR #170, opened 2024-11-28
[th/error-loading-eval-config] evaluator: improve error message when parsing "eval-config.yaml"

thom311:
Previously we only got:

    ValueError: ".IPERF_TCP[0].Normal": mandatory key missing

Now we get:

    ValueError: Failure parsing 'eval-config.yaml': ".IPERF_TCP[0].Normal": mandatory key missing


wizhaoredhat:
LGTM

PR #172, opened 2024-12-03
[th/validate-args] testConfig: improve validation of {ConfServer,ConfClient}.args

thom311:


testConfig.py
```diff
@@ -145,6 +151,14 @@ def _construct_args(pctx2: StructParseParseContext) -> tuple[str, ...]:
 
         return typing.cast("T2", result)
 
+    def _validate(self, test_type: TestType) -> None:
+        if self.args is not None:
+            if test_type not in (TestType.SIMPLE,):
```
wizhaoredhat, line 156: we only support SIMPLE test type?

thom311: this parameter was added especially for "simple", to tweak what the script does. The script is mainly intended for manually hacking it for some ad-hoc testing, so passing additional parameters to it may make sense.

passing args to iperf or netperf tools probably would make sense. But it's not implemented yet, hence it's currently rejected.

wizhaoredhat:
LGTM

PR #173, opened 2024-12-05
[th/generate-eval-config] add new tool for generating eval-config thresholds

thom311:
We had a similar tool, however that bitrot and no longer worked. It was
dropped in [1].

Bring something similar back, but it's also quite different.

The tool now takes a list of previous runs. It computes a new threshold
and writes an "eval-config.yaml".

It can also take an existing configuration with the "--config" option.
In that case it will keep the entries from the configuration and only
update the thresholds. With "--tigthen-only", it will only lower the
thresholds, not increase them.

This is an initial approach to answer [2]. Probably there are still
places to improve later. This is added now, so we have a place for doing
such improvements.

[1] https://github.com/wizhaoredhat/ocp-traffic-flow-tests/commit/137b00dcfb4025ab23011d21cb0860c73da922d2
[2] https://issues.redhat.com/browse/NHE-774

### DIFF
--- a/tftbase.py
+++ b/tftbase.py
@@ -71,6 +71,30 @@ TFT_TESTS = "tft-tests"
 T = typing.TypeVar("T")
 
 
+def eval_binary_opt_in(
+    a: Optional[bool],
+    b: Optional[bool],
+) -> tuple[bool, bool]:
+    if a is None and b is None:
+        # If both are unset, we return True,True
+        return True, True
+
+    # Normalize values to a Optional[bool].
+    if a is not None:
+        a = bool(a)
+    if b is not None:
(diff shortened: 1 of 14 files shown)


tftbase.py
```diff
@@ -71,6 +71,30 @@ def get_tft_image_pull_policy() -> str:
 T = typing.TypeVar("T")
 
 
+def eval_binary_opt_in(
```
wizhaoredhat, line 74: This name is a bit confusing. I am not sure what to call this but does it follow AND/NAND/NOR/OR logic?

wizhaoredhat:
LGTM

PR #176, opened 2024-12-10
[th/print-results-multiple] print_results: support printing multiple results

thom311:
Since "config.yaml" can contain multiple tests and "tft.py" can write multiple result files, so also "print_results.py" should handle multiple files.

That way,

    ./tft.py -o ft-logs/result- config.yaml
    ./print_results.py ft-logs/result-*.json

works to check all result files, regardless how many tests are defined in config.yaml.

### DIFF
--- a/print_results.py
+++ b/print_results.py
@@ -55,8 +55,8 @@ def parse_args() -> argparse.Namespace:
     )
     parser.add_argument(
         "result",
-        type=str,
-        help="The JSON result file from TFT Flow test.",
+        nargs="+",
+        help="The JSON result file(s) from TFT Flow test.",
     )
     common.log_argparse_add_argument_verbose(parser)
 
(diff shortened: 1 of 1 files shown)


thom311:
this is done because in https://gitlab.cee.redhat.com/ccit/jenkins-csb-customers/nst-casc-net-hw-ci/-/blob/5e59d37e2773b6d28ee180490aadf900eb2cbaed/scripts/ovn-traffic-flow/traffic-flow-run-test.jenkinsfile#L59 we do:

```
./print_results.py $(ls ft-logs/*-RESULTS)
```

previously, that only works because the used `config.yaml` happens to only contain one test. This line should always work correctly, regardless how many test results are created by `tft.py`.

wizhaoredhat:
LGTM

PR #184, opened 2024-12-20
[th/less-redundant-args] simplify code about Task's `node_name` and TestCaseType info

thom311:
- Improve `TestCaseTypInfo` which represents the behavior of `TestCaseType`. Also add a unit test that exposes a bug how multiple TestCaseType behave identical (not fixed yet).
- we pass redundant parameters around, for example, the "node_name" argument of the `Task`. I also think, we sometimes set it to the wrong value, because of how confusingly redundant information is used. Fix that by letting the `Task` know its `TaskMode` (client/server). It can thereby determine the right node name based on the configuration (`TestSettings`).


testSettings.py
```diff
@@ -37,9 +35,42 @@ def _post_check(self) -> None:
...
+    @property
+    def conf_client(self) -> testConfig.ConfClient:
+        # For now, only one server/client is supported and TestConfig already
+        # enforces that. Do tuple-unpacking here, to further assert that there
+        # is only one server/client.
+        (c_client,) = self.connection.client
+        return c_client
+
+    @property
+    def conf_server_used(self) -> testConfig.ConfBaseClientServer:
```
wizhaoredhat, line 62: What does `conf_server_used` function name supposed to mean?

Would having NodeName in the function name help with reading? Also it is based on the test case, maybe making that explicit would help people understand the function.

testSettings.py
```diff
@@ -37,9 +35,42 @@ def _post_check(self) -> None:
...
+        if self.test_case_id.info.is_same_node:
+            return self.conf_client
+        else:
+            return self.conf_server
+
+    def conf_clientserver(self, task_role: TaskRole) -> testConfig.ConfBaseClientServer:
```
wizhaoredhat, line 68: Would there be a better name for this? `conf_clientserver` doesn't really tell me what I can expect from this function?

I would expect NodeName as part of the function name,  unless there is a good reason for it to be omitted.

testSettings.py
```diff
@@ -37,9 +35,42 @@ def _post_check(self) -> None:
...
+        # Access some properties here. We would get an exception early on, if
+        # there is something wrong.
         self.connection
-        self.test_case_id
+        self.test_case_id.info
+        self.conf_server
+        self.conf_client
+
+    @property
+    def conf_server(self) -> testConfig.ConfServer:
```
wizhaoredhat, line 46: Where are conf_client and conf_server functions used?

testSettings.py
thom311: it's the effectively used `ConfClient` or `ConfServer`, depending on the `TestCaseType`'s `is_same_node`

testSettings.py
thom311: NodeName doesn't seem good, because it's not the name. It's the configuration of the entire node.

testSettings.py
thom311: at various places. Already previously.

testSettings.py
wizhaoredhat: How about ServerConfByTestCaseType?

testSettings.py
wizhaoredhat: I think these should be internal functions and external users would need to use `conf_server_used`

testSettings.py
wizhaoredhat: How about ClientOrServerConfByTaskRole? WDYT?

testSettings.py
thom311: At the beginning (on `main` branch), the `conf_server` already exists as an attribute and is already in active use. The branch does several patches, making it a property, and then reduce its use. There are still actual users.

Probably there should be more patches done, that renames things and moves things around further. But as far as this branch goes, it's already (too) large and overall goes in the right direction.

I'd rather have this merged, and improve there further. For example, `ConfServer` has an attribute "persistent", but `ConfClient` doesn't have it. Note how `conf_server_used` returns the base class `ConfBaseClientServer`, which also doesn't have a "persistent" flag. Maybe the "persistent" flag is anyway not implemented correctly, but saying just use "conf_server_used" everywhere needs at least address what to do about this flag (e.g. move it to the parent class). The point is, many patches can be done after this PR.

Also, sure, I could not start out and do a major renaming patch first. But that only causes me extra work to rebase this branch on top of that, and it's not clear why the renaming must come before the changes that are anyway planned. The branch itself uses names that make sense on top of what we currently have, although, maybe there should be some renaming done there.

testSettings.py
thom311: renaming should be done. I'd rather do that in a follow-up.

testSettings.py
wizhaoredhat: I see please do a rename in a subsequent PR if you have the cycles. Thank you! We should have descriptive and short names, not just short names for the sake of them being short.

wizhaoredhat: Regarding the "persistent" flag. It was intended for the server only since iperf3 -s can run indefinitely, but iperf3 -c is only one-shot. I think this flag is in the properly place, now regarding whether the "persistent" flag is still working correctly, I am not too sure after all the recent changes. I am guessing most people test with "persistent" flag being false.

The purpose of the "persistent" flag being there is such that the user can run further tests without managing the server.

PR #185, opened 2025-01-07
[th/conf-node-cleanup] rename and cleanup uses of node config

thom311:
This is a follow up to https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/184#discussion_r1897475029


wizhaoredhat:
LGTM

PR #186, opened 2025-01-07
[th/skip-deprecated-aliases-for-test-cases] handle aliases for test cases by skipping

thom311:
Some `TestCaseType` are identical to others. If we select a range (`test_cases: "*"` or `test_cases: "1-26"`), we will effectively run the same test twice. That is undesirable because it wastes time.

Skip those aliases.

An alternative would be to just delete them. But that might break existing configurations. So it's not done (for now).


wizhaoredhat:
I think it would be cleaner to just remove those that don't work or doesn't make sense

thom311:
> I think it would be cleaner to just remove those that don't work or doesn't make sense

ok, how about this?

the two earlier commits are still there, they don't hurt and show how this could have been handled alternatively. I don't mind squashing the branch, if we agree on dropping the values.

wizhaoredhat:
LGTM! Thanks for cleaning this up

PR #188, opened 2025-01-22
[th/kubeconfig-system-check] improve validation of kubeconfig files and add system_check()

thom311:
The kubeconfig files can be part of the configuration YAML. Or they can be detected defaults based on the files we have on the file system. Also, they can be invalid (e.g. not valid YAML).o

Improve the validation of those files, so that we fail early and with a better error message.

Also, validating the files (and resolving the default values) requires to look at the file system. We want to have a first step (the `TestConfig()` constructor) that only parses and validates the configuration file alone. Only a second step would perform further validation (and check the kubeconfig files on disk). Add a `TestConfig.system_check()` for that purpose.

Benefits:

- better error message for invalid kubeconfig (and at a early stage)
- make `TestConfig()` usable to only validate the configuration file, while having `TestConfig.system_check()` to validate further dependencies (e.g. files on disk). With this, we can validate only configuration alone.

### DIFF
--- a/testConfig.py
+++ b/testConfig.py
@@ -563,8 +563,7 @@ class TestConfig:
     configpath: Optional[str]
     configdir: str
     cwddir: str
-    kubeconfig: str
-    kubeconfig_infra: Optional[str]
+    _kubeconfig_pair: Optional[tuple[str, Optional[str]]]
     _client_tenant: Optional[K8sClient]
     _client_infra: Optional[K8sClient]
     _lock: threading.Lock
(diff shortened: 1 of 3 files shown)


testConfig.py
```diff
@@ -655,70 +667,78 @@ def __init__(
...
+        self._lock = threading.Lock()
         self._client_tenant = None
         self._client_infra = None
 
         if not output_base:
             output_base = None
         self.output_base = output_base
 
-        kubeconfigs_cwd = cwddir
+        kubeconfig_pair1: tuple[Optional[str], Optional[str]]
```
wizhaoredhat, line 678: why is it `kubeconfig_pair1` and not `kubeconfig_pair`

thom311: there is already a variable `kubeconfig_pair` (the argument to this function).

wizhaoredhat:
LGTM

PR #190, opened 2025-02-03
Remove workaround for resource limits and requests in sriov pod yaml

vrindle:
The reason for this change is that when the PR for enabling the SRIOV dp admission controller goes in, the resource requests and limits should be injected automatically without specification in the yaml. Therefore these traffic flow tests need to be updated to reflect that change.

### DIFF
--- a/manifests/sriov-pod.yaml.j2
+++ b/manifests/sriov-pod.yaml.j2
@@ -17,9 +17,4 @@ spec:
     image: {{ test_image }}
     command: {{ command }}
     args: {{ args }}
-    imagePullPolicy: {{ image_pull_policy }} {% if use_secondary_network and resource_name %}
-    resources:
-      requests:
-        {{ resource_name }}: '1'
-      limits:
-        {{ resource_name }}: '1' {% endif %}
+    imagePullPolicy: {{ image_pull_policy }}


thom311:
> PR for enabling the SRIOV dp admission controller goes in

Could you link to that PR?

vrindle:
https://github.com/openshift/dpu-operator/pull/172

thom311:
is the `PodType.SRIOV` (the `"sriov"` key in the configuration YAML) only relevant with dpu-operator? What if we use sriov-network-operator, is this change then relevant, and how does it affect it?

> FYI... This PR in the traffic flow test tool should go in before the one I just linked

Your patch deletes explicitly setting the limits, presumably because the mutating admission webhook will add them. But then, if the user keeps adding them explicitly, it's still going to work, isn't it? If so, why does this PR need to go in first, and is it strictly speaking *necessary*? Because, this change will also break tft tool with older versions of dpu-operator that don't have the webhook, will it not?

vrindle:
> is the `PodType.SRIOV` (the `"sriov"` key in the configuration YAML) only relevant with dpu-operator? What if we use sriov-network-operator, is this change then relevant, and how does it affect it?

I don't think this is only for the dpu-operator. However the sriov-network-operator already has a network-resources-injector so it should not affect it (keep in mind this change is reverting the flow test tool before we implemented the workaround)

> 
> > FYI... This PR in the traffic flow test tool should go in before the one I just linked
> 
> Your patch deletes explicitly setting the limits, presumably because the mutating admission webhook will add them. But then, if the user keeps adding them explicitly, it's still going to work, isn't it? If so, why does this PR need to go in first, and is it strictly speaking _necessary_? Because, this change will also break tft tool with older versions of dpu-operator that don't have the webhook, will it not?

The reason this PR needs to go in is because this PR is how we test if the PR I linked passes it's end to end tests. Once this one goes in, I will bump the ocp-traffic-flow tests commit in that PR and the pods should be created. Removing this PR is necessary to adequately test e2e changes in the linked PR. The yaml was only supposed to be temporary until the webhook was complete.

wizhaoredhat:
I think NVIDIA contributed to this YAML as well for secondary networks. Maybe the compromise is to point this tool to custom YAML manifest files. Like a option to change the default directory to look for the manifests YAML file. WDYT?

thom311:
> I think NVIDIA contributed to this YAML as well for secondary networks.

NVIDIA contributed tests 27-29. I am not sure how this change affects that.

This is the problem here. It's not clear who runs the tests in which exact setup. Optimally, we have CI for every setup we care about. In practice, we don't want to invest that effort.

That's why effort should to be taken to understand how various (loosely-defined) setups will be affected by the change.

thom311:
@vrindle I am really mostly fine with it. Please improve the commit message and we are IMO good.

vrindle:
@thom311 and I discussed his concerns over video call. I added additional context and he has agreed that the commit message regarding why this change is occurring can stay. We agreed to add additional info to the commit message regarding the network resources injector being necessary for running tests (should not disable).

wizhaoredhat:
So the code will do this

If (directory provided) {
    If (yaml file found matched by name) {
         Use this yaml file
         Return
    }
}
Use default yaml file in default location

vrindle:
@wizhaoredhat I’m still a bit unclear on why we’re keeping this workaround in place. When we initially added it, we agreed it would be a temporary solution and would be reverted once the webhook PR was completed.

vrindle:
@wizhaoredhat @thom311 one thing we can also do is to remove references to secondary_nad and resource_name in the ocp_tft config in the dpu-operator instead of this change... that way the dpu-operator can still be tested properly and we can still keep this workaround for when webhooks are disabled.

vrindle:
I added the commit here to disable without removing the workaround... this should be good enough (https://github.com/openshift/dpu-operator/pull/172/commits/f50382b57a860d5c7fc118b2239e313889bee152#diff-8182782e714f517eba6650339311a6ec39a8d1b6ad057270150cf68d630b0d34L22). We could close this PR...

wizhaoredhat:
> I added the commit here to disable without removing the workaround... this should be good enough (https://github.com/openshift/dpu-operator/pull/172/commits/f50382b57a860d5c7fc118b2239e313889bee152#diff-8182782e714f517eba6650339311a6ec39a8d1b6ad057270150cf68d630b0d34L22). We could close this PR...

Did you test it? I don't think this works....

vrindle:
@wizhaoredhat I haven't tested it yet... but this should work. Why don't you think it works?

wizhaoredhat:
> @wizhaoredhat I haven't tested it yet... but this should work. Why don't you think it works?
> 

PTL at

  annotations: {% if not use_secondary_network %}
    v1.multus-cni.io/default-network: {{ default_network }} {% else %}
    k8s.v1.cni.cncf.io/networks: {{ secondary_network_nad }} {% endif %}

vrindle:
@wizhaoredhat aah I see... it should be fine to just disable resource name then...

thom311:
> it should be fine to just disable resource name then and keep network attachment definition

I don't think so. If you omit the resource_name, then it gets automatically detected. See [here](https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/blob/main/task.py#L342). That is also what the README says:

> This field is optional and will default to None if not set, but if secondary network nad is defined, traffic flow test
tool will try to autopopulate resource_name based on the secondary+network_nad provided.

vrindle:
Ahh It does look like it autopopulates it... I forgot I added that in...

PR #191, opened 2025-02-05
[th/manifests-overrides] tft: support manifests/overrides for providing alternative manifest files

thom311:
Under "manifests" directory we have various files that we render.  Add a mechanism so that the user can provide alternative files.

The user can set TFT_MANIFEST_OVERRIDES environment variable, which can be an overrides directory. If unset, it defaults to "manifests/overrides/". If empty, no overrides are used.

If tft looks for manifest files to render (like "pod.yaml.j2"), it now first looks at the overrides directory. If the file exists, it is taken. It shadows the file that we commit to git. Otherwise, tft takes the file from "manifests/" directory, as before.

---

Follow-up to https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/190
CC @wizhaoredhat @vrindle


wizhaoredhat:
lgtm

vrindle:
lgtm

PR #192, opened 2025-02-06
[th/manifests-yamls] improve handling manifests/yamls directory for rendering YAMLs

thom311:
- fix handling empty `TFT_MANIFESTS_OVERRIDES=`.
- cleanup `get_tft_manifests_overrides()` and log value
- improve handling of "manifests/yamls" output to not rely on current working directory.
- make "manifests/yamls" directory configurable via `$TFT_MANIFESTS_YAMLS`

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,3 +121,5 @@ Simply run the python application as so:
 - `TFT_MANIFESTS_OVERRIDE` to specify an overrides directory for manifests. If not set, the
      default is "manifests/overrides". If set to empty, no overrides are used. You can place
      your own variants of the files from "manifests" directory and they will be preferred.
+- `TFT_MANIFESTS_YAMLS` to specify the output directory for rendered manifests. This
+     defaults to "manifests/yamls".
--- a/pluginMeasureCpu.py
+++ b/pluginMeasureCpu.py
@@ -53,8 +53,8 @@ class TaskMeasureCPU(PluginTask):
         )
 
         self.in_file_template = tftbase.get_manifest("tools-pod.yaml.j2")
-        self.out_file_yaml = (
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -121,3 +121,5 @@ Simply run the python application as so:
 - `TFT_MANIFESTS_OVERRIDE` to specify an overrides directory for manifests. If not set, the
```
wizhaoredhat, line 121: should this be env as input or should we use command line arguments?

thom311: maybe. I think env variables are an actually very nice way for configuring things.

For example, it gets passed across processes (e.g. dpu-operator's `make e2e-tests` wouldn't expose a command line option for this, but `TFT_MANIFESTS_OVERRIDE=/path make e2e-tests` would work as one expects.

I like configuration via environment variables, but I guess, some people dislike it.

wizhaoredhat: fair enough argument. Let's keep as is then

wizhaoredhat:
lgtm

README.md
thom311: @wizhaoredhat also,

I think programs should almost never call `setenv()` (except between `fork()`+`exec()`). That means, the environment is immutable.

As such, if `setenv()` is not called, then the function 
```
@functools.cache
def get_tft_manifests_yamls() -> str:
   ...
```
can be called at any time.

If the parameter is configurable via command line, then first we must call `argparse.parse()`, and somehow propagate the value over to `get_tft_manifests_yamls()` (like having an `initialize()` function). And we cannot call get_tft_manifests_yamls() before it was initialized. This could be handled, for example to `get_tft_manifests_yamls()` raise an exception when being called without being initialized. But it's extra effort. With environment variables it's nicer.

PR #194, opened 2025-02-20
[th/aggregate-result-cleanup] cleanup Task.aggregate_output()

thom311:
- better validate the result when we receive it. This way, we ensure that the result dictionary contains what we expect, or we set `success=False` and a `msg` reason.
- move common code to the `Task.aggregate_output()` parent implementation.

### DIFF
--- a/pluginMeasureCpu.py
+++ b/pluginMeasureCpu.py
@@ -1,8 +1,11 @@
 import logging
 import jc
+import typing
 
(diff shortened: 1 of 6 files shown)


wizhaoredhat:
@thom311 please fix one of the commits that doesn't have a signoff.

wizhaoredhat:
LGTM

PR #199, opened 2025-03-19
[th/render-undefined-variables] task: rework setting template variables by moving to parent Task.get_template_args()

thom311:
With a recent change in ktoolbox.kjinja2.render_data(), the function now errors out (by default) if an undefined variable is referenced. That is useful, to find bugs about undefined variables in Jinja2 templates.

There are known cases where we render manifest files for client pods, that refer to `{{port}}` jinja2 variable, but `Task.get_template_args()` does not set that variable. That leads to a crash. Fix that.

But also, let the parent implementation in `Task.get_template_args()` set more of those variables. Instead of subclasses overwriting `get_template_args()`, the Task parent class is already aware of most of those parameters (`pod_name`, `port`, `test_image`, `default_network`). Let the parent class always set those variables, so that subclasses don't need to. This means we less carefully try to set variables only for manifest templates that need them, but rather define common variables always (in the base class). The benefit is that we need to less carefully estimate when a variable is needed, and just always set it at one place.

Also, simplify constructing the `cmd_line_args()` for `ServerTask` subclasses.

### DIFF
--- a/task.py
+++ b/task.py
@@ -332,7 +332,7 @@ class Task(ABC):
             "test_image": tftbase.get_tft_test_image(),
             "image_pull_policy": tftbase.get_tft_image_pull_policy(),
             "command": ["/usr/bin/container-entry-point.sh"],
-            "args": [],
+            "args": self._get_template_args_args(),
             "index": f"{self.index}",
             "node_name": self.node_name,
             "pod_name": self.pod_name,
@@ -355,6 +355,9 @@ class Task(ABC):
 
     def _get_template_args_port(self) -> str:
         return ""
+
+    def _get_template_args_args(self) -> list[str]:
+        return []
 
     def render_file(
         self,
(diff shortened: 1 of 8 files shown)


task.py
```diff
@@ -255,6 +255,7 @@ def __init__(
         index: int,
         tenant: bool,
         task_role: TaskRole,
+        port: int = 35201,
```
wizhaoredhat, line 258: I kind of don't like the bogus port number. It might be tough to spot and debug in the future. Can it by default be a string like "N/A" or "None".

thom311: I choose this bogus number, because it's also unique.

The debug logs of TFT contain the rendered YAML files (and of course, you can find the rendered files on disk after the test -- though, if you run multiple cases they overwrite each other).

Anyway. If you see `35201`, then you can git-grep for that, and you will only find this one occurrence in the sources. Also, you can easily review that of the various subclasses, only `PluginTask` (and `PluginTask`'s subclasses) do not set the `port` in the constructor. So it know where such value comes from.

In practice, you would not see the value, because we simply don't render templates for cases where the port is left at the 35201 default. If you really end up seeing this number somewhere, then there is a bug that should be fixed (or you injected a TFT_MANIFESTS_OVERRIDES template with a `{{port}}` for a situation that has no port -- which is also wrong usage). In either case, it's easy to git-grep and understand where this went wrong (and fix it).

I don't like 35201 very much too. But I like it better than a string "N/A" or "None", because currently `Task.port` is an `int`. I could make it a `Optional[int]`, but then I need to specially handle None at various places...

thom311: ok, I reworked the branch significantly. I think it's better now.

wizhaoredhat:
I think you need to revise the PR and commit msg?

thom311:
setting to Draft for now. Seems the ovn-kubernetes tests are failing. Investigating why that is (and whether that is caused by this PR).

PR #202, opened 2025-04-05
docs: add developer setup instructions to README

SamD2021:
Adds a section explaining how to install development dependencies using requirements-dev.txt, which includes both runtime and developer tools like pytest, black, mypy, and flake8. This improves onboarding for contributors by clarifying how to set up a local dev environment.

This step is optional and does not affect script users.


thom311:
Merged. Thank you.

PR #204, opened 2025-04-10
[th/simple-exec] add options to improve "simple" test

thom311:
The "simple" test is supposed to be the hackable one, that you can use for debugging an issue.

For example, with "--exec", it can download arbitrary scripts and exec those instead.

Some improvements to make that better and more usable.


wizhaoredhat:
LGTM

PR #206, opened 2025-04-11
[th/log-duration] log summary line including test duration

thom311:
it's useful to see a line indicating success/failure. Also, show how long the test was running.

while you can figure that out via other means (e.g. look at timestamps from the logs), this should be visible at one glance.


wizhaoredhat:
LGTM

PR #208, opened 2025-04-15
[th/simple-exec-improvement] simple: improve "simple-exec.sh" script

thom311:
By passing an argument to "simple-exec.sh" (e.g. via the "--exec-arg" option), the script downloads another "simple-tcp-server-client.py" to run instead.

That is useful, because we might still want to run something like the "simple-tcp-server-client.py" script, but the one inside in the container is more cumbersome to update during a quick test.


wizhaoredhat:
LGTM

PR #209, opened 2025-04-15
[th/privileged-pod] task: support configuring test pods as priviledged

thom311:
Usually we don't need this, and usually we want to run with restricted permissions to be closer to what the user would do.

However, for testing it can be useful (e.g. to exec tcpdump in a running container). Make it configurable via configuration YAML and `TFT_PRIVILEGED_POD` environment variable.

### DIFF
--- a/task.py
+++ b/task.py
@@ -336,6 +336,10 @@ class Task(ABC):
             "node_name": self.node_name,
             "pod_name": self.pod_name,
             "port": self._get_template_args_port(),
+            "privileged_pod": common.bool_to_str(
+                self._get_template_args_privileged_pod(),
+                format="true",
+            ),
             "secondary_network_nad": self.ts.connection.effective_secondary_network_nad,
             "use_secondary_network": (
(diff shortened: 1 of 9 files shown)


wizhaoredhat:
LGTM

task.py
```diff
@@ -352,6 +356,18 @@ def get_template_args(self) -> dict[str, str | list[str]]:
...
+    def _get_template_args_privileged_pod(self) -> bool:
+        v = self.node.privileged_pod
+        if v is not None:
+            # The per-node setting has precedence.
+            return v
+        v = tftbase.get_tft_privileged_pod()
+        if v is not None:
+            # Then the setting from the environment variable.
+            return v
+        # Finaly, the test-wide setting fro the configruration file.
```
wizhaoredhat, line 368: Some typos here.

thom311: https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/212

PR #218, opened 2025-05-22
Sanitize node name before creating pod name

m-naught:
Pods can't contain dots, replace them with dashes if the node name has them.

### DIFF
--- a/pluginMeasureCpu.py
+++ b/pluginMeasureCpu.py
@@ -54,11 +54,9 @@ class TaskMeasureCPU(PluginTask):
             tenant=tenant,
         )
 
+        self.pod_name = f"tools-pod-{self.node_name_sanitized()}-measure-cpu"
         self.in_file_template = tftbase.get_manifest("tools-pod.yaml.j2")
-        self.out_file_yaml = tftbase.get_manifest_renderpath(
-            f"tools-pod-{self.node_name}-measure-cpu.yaml"
-        )
-        self.pod_name = f"tools-pod-{self.node_name}-measure-cpu"
+        self.out_file_yaml = tftbase.get_manifest_renderpath(self.pod_name + ".yaml")
 
     def initialize(self) -> None:
         super().initialize()
--- a/task.py
+++ b/task.py
@@ -291,6 +291,9 @@ class Task(ABC):
     def node_name(self) -> str:
         return self.node.name
 
+    def node_name_sanitized(self) -> str:
+        return self.node_name.replace(".", "-")
+
     def get_namespace(self) -> str:
         return self.ts.cfg_descr.get_tft().namespace
 
@@ -753,7 +756,7 @@ class ServerTask(Task, ABC):
 
         connection_mode = ts.connection_mode
         pod_type = ts.server_pod_type
-        node_name = self.node_name
+        node_name_sanitized = self.node_name_sanitized()
         port = 5201 + self.index
 
         if connection_mode == ConnectionMode.EXTERNAL_IP:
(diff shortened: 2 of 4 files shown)


task.py
```diff
@@ -291,6 +291,15 @@ def node(self) -> testConfig.ConfNodeBase:
     def node_name(self) -> str:
         return self.node.name
 
+    def node_name_sanitized(self) -> str:
```
thom311, line 294: typo in commit message.

Also, please explain the use case for this. The name comes from the YAML configuration. Is somebody configuring names with dots? Why?

It seems confusing that the user configures some name, and then that name is not usable and instead we remap it to something else. Why not
```diff
diff --git i/testConfig.py w/testConfig.py
index 481242ed7649..529463bfba07 100644
--- i/testConfig.py
+++ w/testConfig.py
@@ -100,6 +100,9 @@ class ConfNodeBase(_ConfBaseConnectionItem, abc.ABC):
 
             name = common.structparse_pop_str_name(varg.for_name())
 
+            if "." in name:
+                raise pctx.value_error("name cannot contain dots")
+
             sriov = common.structparse_pop_bool(
                 varg.for_key("sriov"),
                 default=False,
```

task.py
```diff
@@ -294,12 +294,11 @@ def node_name(self) -> str:
     def node_name_sanitized(self) -> str:
         st = list(self.node.name)
         for i, _ in enumerate(st):
-            if st[i] == '.':
-                st[i] = '-'
+            if st[i] == ".":
+                st[i] = "-"
```
thom311, line 298: let's not do a commit that fixes  earlier commits of the same work.

Every commit individually should be sensibly correct, self-contained and pass basic checks (in this case `black` formatting).

task.py
thom311: or maybe better, add a `validate_regex: Optional[Union[str, re.Pattern]] = None` argument to `structparse_pop_str_name` and call

```python
name = common.structparse_pop_str_name(varg.for_name(),
                                       validate_regex="^[-_a-zA-Z0-9]+$")
```

m-naught: There already is an error if the pod name contains dots, hence this patch to circumvent it.

This change was needed when I was testing for the e825c card. The node name was just a hostname, which I couldn't change.

thom311: > The node name was just a hostname, which I couldn't change.

Why you cannot change that? Where is that configuration file?

m-naught: I was not managing the cluster

thom311: Still, please explain this in the commit message. Pointing out the thing that seems obvious to you, could have avoided this discussion :) .

Basically:

```
The node name primarily refers to the hostname of the node and can contain dots.
When generating a pod name from that name, the dots must be sanitized.
```

task.py
```diff
@@ -920,7 +928,7 @@ def __init__(self, ts: TestSettings, server: ServerTask):
         )
 
         pod_type = ts.client_pod_type
-        node_name = self.node_name
+        node_name = self.node_name_sanitized()
```
thom311, line 931: please rename the variable `node_name` to `node_name_sanitized`, as this is what it is now.

pluginMeasureCpu.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, ts: TestSettings, task_role: TaskRole, tenant: bool):
         self.out_file_yaml = tftbase.get_manifest_renderpath(
             f"tools-pod-{self.node_name}-measure-cpu.yaml"
```
thom311, line 59: If you give the pod another name, then the yaml file should also follow (and use the sanitized name).

task.py
```diff
@@ -291,6 +291,9 @@ def node(self) -> testConfig.ConfNodeBase:
     def node_name(self) -> str:
         return self.node.name
 
+    def node_name_sanitized(self) -> str:
+        return self.node.name.replace(".", "-")
```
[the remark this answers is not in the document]
thom311: anyway. That can be done separately. Let's merge this.

PR #231, opened 2025-10-31
Add a wait for listen socket before IperfClient connects

wizhaoredhat:
We should make sure that the IperfServer is listening on the tcp/udp socket first before allowing the IperfClient to connect.
    
This makes sure that we don't get into any race conditions with a slow start up of iperf3 -s.

### DIFF
--- a/testTypeIperf.py
+++ b/testTypeIperf.py
@@ -125,6 +125,13 @@ class IperfServer(task.ServerTask):
     def _create_setup_operation_get_cancel_action_cmd(self) -> str:
         return f"killall {IPERF_EXE}"
 
+    def _get_server_listen_protocol(self) -> Optional[str]:
+        # You may wonder why even for TestType.IPERF_UDP, we would detect for TCP here.
+        # The reason is that iperf3 server always listens on TCP even for UDP tests.
+        # Iperf3 defaults to listening to TCP port for the control channel. This is used for
+        # managing the test setup and reporting.
+        return "tcp"
+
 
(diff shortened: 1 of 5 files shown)


testTypeSimple.py
```diff
@@ -48,6 +48,9 @@ def cmd_line_args(self, *, for_template: bool = False) -> list[str]:
     def _create_setup_operation_get_cancel_action_cmd(self) -> str:
         return "killall python3"
 
+    def _get_server_listen_protocol(self) -> str:
+        return "tcp"
```
thom311: I was wondering, whether the parent implementation should default to `return "tcp"`. Then most subclasses wouldn't need to override the function, it also seems a good default for a "Server"Task.

If we ever add a future `ServerTask` that does not have an actual port, then that one can override the function to `return None`.

testTypeIperf.py
```diff
@@ -125,6 +125,11 @@ def cmd_line_args(self, *, for_template: bool = False) -> list[str]:
     def _create_setup_operation_get_cancel_action_cmd(self) -> str:
         return f"killall {IPERF_EXE}"
 
+    def _get_server_listen_protocol(self) -> str:
```
thom311: this makes mypy sufficiently happy (for good reasons).

I still think that the overwritten function signature should be `_get_server_listen_protocol(self) -> Optional[str]:` to match the parent's signature. Even if our implementation in the subclass never actually returns `None`...

wizhaoredhat: done

PR #247, opened 2026-02-26
Add EgressIP support as connection mode parameter

jkary:
## Summary
- Add `POD_TO_EXTERNAL_EGRESS` test case type for EgressIP testing
- Add `egress_ip` as a connection parameter with IP validation and optional node selection
- Add `create_egressip()` to Task class following the ANP resource creation pattern
- Add EgressIP CR and node label cleanup in `_cleanup_previous_testspace()`

### Usage
```yaml
connections:
  - name: "test"
    type: iperf-tcp
    egress_ip:
      ip: "192.168.1.100"
      node: "worker-1"  # optional, defaults to client node
    external_server_ip: "10.0.0.1"  # optional
```

## Test plan
- [x] Unit tests pass (`pytest` — 42 tests)
- [x] Pre-commit hooks pass (black, flake8, mypy)
- [x] Verified with live EgressIP test against OVN-Kubernetes cluster
- [x] Rebased onto origin/main
- [x] Addressed review feedback:
  - Removed egress-specific properties from `testSettings.py`
  - Consolidated 6 helper methods into single `create_egressip()` matching ANP pattern
  - Decoupled from DPU — uses `self.run_oc()` throughout
  - CR and node label cleanup confirmed in `trafficFlowTests.py`

### DIFF
--- a/task.py
+++ b/task.py
@@ -1,4 +1,5 @@
 import enum
+import ipaddress
 import json
 import logging
 import os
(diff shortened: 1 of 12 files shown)


tftbase.py
```diff
@@ -321,7 +317,7 @@ class TestCaseType(Enum):
     HOST_TO_NODE_PORT_TO_HOST_SAME_NODE = 23
     HOST_TO_NODE_PORT_TO_HOST_DIFF_NODE = 24
     POD_TO_EXTERNAL = 25
-    HOST_TO_EXTERNAL = 26
+    POD_TO_EXTERNAL_EGRESS = 26
```
aserdean: I think we still need to enhance the PR:

multiple egress IPs, secondary host NIC egress. it can be the subject of this PR or subsequent ones.

We do need to think if we would also want to do validation:
i.e. failover mid-stream, no source IP validation

WDYT @wizhaoredhat @tsorya ?

task.py
```diff
@@ -785,6 +786,205 @@ def pod_get_vf_rep(
...
+                ifaddr = json.loads(result.out.strip().strip("'\""))
+                if "ipv4" in ifaddr:
+                    # Extract CIDR from address like "192.168.12.126/24"
+                    addr_with_prefix = ifaddr["ipv4"]
+                    network = ipaddress.ip_network(addr_with_prefix, strict=False)
+                    return (node_ip, str(network))
+            except (json.JSONDecodeError, KeyError, ValueError) as e:
+                logger.warning(f"Failed to parse node-primary-ifaddr: {e}")
+
+        # Fallback: assume /24 network
```
tsorya: Should we have fallback at all? This is test, imo scenario should be super clear

tftbase.py
tsorya: I believe we can enhance it as with subsequent prs, just need to open tickets for it

testSettings.py
```diff
@@ -94,6 +94,26 @@ def initialize_clmo_barrier(self, parties: int) -> None:
     def connection(self) -> testConfig.ConfConnection:
         return self.cfg_descr.get_connection()
 
+    @property
+    def has_egress_ip(self) -> bool:
```
tsorya: same problem as in previous PR, this is something specific to Egress test case, why we need it here?

task.py
```diff
@@ -784,6 +786,177 @@ def pod_get_vf_rep(
...
+        # Format: {"ipv4":"192.168.12.126/24","ipv6":"..."}
+        try:
+            ifaddr = json.loads(result.out.strip().strip("'\""))
+        except json.JSONDecodeError as e:
+            raise json.JSONDecodeError(
+                f"Failed to parse node-primary-ifaddr annotation on node {egress_node}",
+                e.doc,
+                e.pos,
+            ) from e
+        if "ipv4" not in ifaddr:
```
tsorya:
```suggestion
        if "ipv4" not in ifaddr:
```try:
    value = ifaddr["ipv4"]
except KeyError:
   log...
    raise

josecastillolema:
Hi!
Have just tested this PR, I believe it is missing validation that the traffic actually uses the EgressIP as source.
Correct me if I am wrong, but `create_egressip()` method (lines 833-888):
1. Validates the node exists
2. Validates the EgressIP is in the node's subnet
3. Labels the node as egress-assignable
4. Renders and applies the EgressIP CR
5. Does a get egressip only to confirm the CR exists, not to check `.status`

There's no verification that:
- The EgressIP was actually assigned (`.status.items` populated)
- OVN programmed the NAT rule
- The pod's egress traffic uses the EgressIP as source

wizhaoredhat:
^ @josecastillolema  has taken over the Egress IP feature and is merged. Closing the PR.

PR #278, opened 2026-04-09
Catch pre_provision errors and clean up

Venkat-Kunaparaju:
Update to cleanup resources during pre_provisioning if an error occurs. Previously, if an error occurred during pre_provisioning then all the resources would hang.


wizhaoredhat:
LGTM

PR #279, opened 2026-04-09
Use test duration for curl and iperf connect timeouts

Venkat-Kunaparaju:
Replace hardcoded timeout values with the configured test duration so timeouts scale with the test configuration. Updated to timeout regardless of expected block or not.


wizhaoredhat:
LGTM

PR #280, opened 2026-04-09
Suppress cleanup errors to debug log level

Venkat-Kunaparaju:
Update error messages (podman, adminnetworkpolicies, missing secondary network annotation, etc.) from ERROR/WARN to DEBUG so they don't clutter normal test output.

Update podman cleanup to succeed if podman does not exist locally, meaning that it is not applicable.


wizhaoredhat:
LGTM

PR #281, opened 2026-04-09
Exclude plugins from pre-provisioning

Venkat-Kunaparaju:
Server and client tasks have standardized initialize functions that safely skip re-initialization when pods are already deployed in pre-provision mode. Plugins however are flexible and their initialize functions may depend on test-specific state that is only available on the instance running the test, not the instance created during provisioning. Since provisioned plugin objects are discarded before the test run, the test-run instances never have their data populated.

To avoid this, plugins are no longer pre-provisioned. They are always created, initialized, and deployed per-test, ensuring initialize always runs on the actual test instance.


wizhaoredhat:
LGTM

PR #282, opened 2026-04-09
Remove port from pod names and use pod name for service selectors

Venkat-Kunaparaju:
Pod names no longer include the port suffix, allowing pods to be reused across test types in pre_provision mode. Service selectors now match on tft-pod-name instead of tft-port and tft-network. Reduces the number of overall pods deployed if running with different connection types concurrently and the number of pods will stay consistent as connection types increase.

Created ensure_services() from initialize() so pre-provisioned server pods create services, which are still unique to the test type since they operate on port-protocol binding.

### DIFF
--- a/manifests/host-pod.yaml.j2
+++ b/manifests/host-pod.yaml.j2
@@ -5,7 +5,7 @@ metadata:
   name: {{ pod_name }}
   labels:
     tft-tests: {{ label_tft_tests }}
-    tft-port: "{{ port }}"
+    tft-pod-name: {{ pod_name }}
     tft-network: primary
 spec:
   hostNetwork: true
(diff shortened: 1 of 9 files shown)


Venkat-Kunaparaju:
Ran tests 1-24,27-31 with both iperf and curl enabled and had 9 active pods:

host-pod-diff-node-client
host-pod-same-node-client
host-pod-server
normal-pod-diff-node-client
normal-pod-same-node-client
normal-pod-secondary-diff-node-client
normal-pod-secondary-same-node-client
normal-pod-secondary-server
normal-pod-server

wizhaoredhat:
LGTM

PR #283, opened 2026-04-09
Add debug logging for test failures and suppress expected block errors

Venkat-Kunaparaju:
Add debug-level context (pod name, port, node, connection mode) to task failure, server startup, and server listen timeout paths for easier tracing since in #282 port will no longer be in pod names. Suppress curl and iperf failure messages for expects_blocked deny tests by adding to debug level logging.


wizhaoredhat:
LGTM

PR #284, opened 2026-04-10
Omit bitrate and reverse fields from HTTP test results

Venkat-Kunaparaju:
HTTP tests do not measure throughput so TX/RX Bitrate and Reverse fields were always printed as None. Hide them for HTTP to reduce length of the test summary output.


wizhaoredhat:
LGTM

PR #285, opened 2026-04-15
Skip unnecessary secondary NAD creation

Venkat-Kunaparaju:
When server/client nodes already have NADs specified, skip creating the fallback tft-secondary NAD. Also exclude UDN test cases from triggering NAD creation in the ft namespace, since UDN tests run in ft-udn with their own NAD setup.


wizhaoredhat:
LGTM

PR #287, opened 2026-04-22
Support for 2nd interface validate_offload

Venkat-Kunaparaju:
eth0 is hardcoded as the interface for offload validation which is used for the primary network.

Updated to support net1 and determine which interface to use depending on the test type.


wizhaoredhat:
LGTM

PR #290, opened 2026-04-27
Add UDN_PRIMARY external, NP, and load balancer test cases

Venkat-Kunaparaju:
Extends the UDN_PRIMARY block with new test cases to support NetworkPolicy, LoadBalancer, and External test types.

Cleanup is extended to the UDN namespace so NetworkPolicy and LoadBalancer Service resources created in {ns}-udn are deleted properly. Centralize the UDN namespace convention in a new function and call it whenever udn namespace is required.


wizhaoredhat (approved):
LGTM

wizhaoredhat:
@Venkat-Kunaparaju  later on we need to stabilize the test case number ID for backwards compatibility. I think especially when we create our first release.

PR #291, opened 2026-04-28
Move validateOffload plugin full results to DEBUG level

Venkat-Kunaparaju:
Currently a lot of unneeded packet statistics are reported at the INFO level. Moving to debug, but keeping rx/tx_packet_start and rx/tx_packet_end in INFO level, along with success boolean and message.


wizhaoredhat (approved):
LGTM

PR #292, opened 2026-05-06
Run HTTP test curl once and use duration as timeout directly

Venkat-Kunaparaju:
Drop the retry loop in HTTP client task that runs for the specified duration. Invoke curl once and the result for it determines the test outcome. Used the exact duration value for the curl timeout.

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP test behavior refined: tests that expect blocking now poll until completion or timeout, while non-blocking tests run once—reducing unnecessary retries and improving speed.
  * Result evaluation preserved, so pass/fail reporting and timeout handling remain consistent.

### DIFF
--- a/testTypeHttp.py
+++ b/testTypeHttp.py
@@ -97,16 +97,18 @@ class HttpClient(task.ClientTask):
             else:
                 _check_success = _check_success_podman
 
-            sleep_time = 0.2
-            end_timestamp = time.monotonic() + self.get_duration() - sleep_time
-
-            while True:
+            if expects_blocked:
+                sleep_time = 0.2
+                end_timestamp = time.monotonic() + self.get_duration() - sleep_time
+                while True:
+                    r = self.run_oc_exec(cmd, may_fail=expects_blocked)
+                    if not _check_success(r):
+                        break
+                    if time.monotonic() >= end_timestamp:
+                        break
+                    time.sleep(sleep_time)
+            else:
                 r = self.run_oc_exec(cmd, may_fail=expects_blocked)
-                if not _check_success(r):
-                    break
-                if time.monotonic() >= end_timestamp:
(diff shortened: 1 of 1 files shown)


testTypeHttp.py
```diff
@@ -97,16 +95,7 @@ def _thread_action() -> BaseOutput:
...
-            end_timestamp = time.monotonic() + self.get_duration() - sleep_time
-
-            while True:
-                r = self.run_oc_exec(cmd, may_fail=expects_blocked)
-                if not _check_success(r):
-                    break
-                if time.monotonic() >= end_timestamp:
-                    break
-                time.sleep(sleep_time)
+            r = self.run_oc_exec(cmd, may_fail=expects_blocked)
```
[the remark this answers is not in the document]
Venkat-Kunaparaju: For the allow case, we already confirm pod is up and server is alive. We also have connection-timeout value to make sure endpoint reachable. 

For the deny case, will update to preserve the old behavior.

testTypeHttp.py
```diff
@@ -97,16 +97,18 @@ def _thread_action() -> BaseOutput:
             else:
                 _check_success = _check_success_podman
 
-            sleep_time = 0.2
-            end_timestamp = time.monotonic() + self.get_duration() - sleep_time
-
-            while True:
+            if expects_blocked:
```
wizhaoredhat, line 100: We might need to change some plugin like measure_power. 
Because the tool expects that duration would be honored.
In this case there is only 1 iteration for HTTP. This won't work for some plugins.

wizhaoredhat: Nevermind, this is a non issue because we won't use measure_power for HTTP tests.

wizhaoredhat (approved):

PR #296, opened 2026-05-27
Fix primary pod IP selection for shared secondary pods

Venkat-Kunaparaju:
Normal pods shared by pre_provision can have a secondary NAD attached, but primary-network test cases should still target the pod's primary IP. Use the secondary IP decision helper for pod IP lookup so secondary IPs are only selected when the test traffic requires them.

Keep SR-IOV behavior unchanged by treating configured secondary NADs as secondary-IP traffic for SR-IOV pods.

### DIFF
--- a/task.py
+++ b/task.py
@@ -487,11 +487,15 @@ class Task(ABC):
 
     @property
     def uses_secondary_ip(self) -> bool:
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
LGTM

PR #297, opened 2026-05-28
Add service target access subtests

Venkat-Kunaparaju:
Add a target access mode for service-backed tests so ClusterIP and LoadBalancer can exercise service DNS names, and NodePort can also target the server node IP. Keep the legacy IP path as the default unless TFT_ENABLE_TARGET_ACCESS_SUBTESTS is enabled.

Carry the selected target access mode through test settings and metadata, while omitting the default IP mode from serialized results and summary output to preserve existing result formats.

## Summary by CodeRabbit

* **New Features**
  * Added configurable target access modes so tests can run each connection instance across supported modes (service DNS/IP variants, and additional NodePort/LoadBalancer coverage when enabled).
  * Test output now shows a “Target Access” line when the selected mode isn’t the computed default.
* **Tests**
  * Target access mode is carried through test settings and metadata end-to-end.
  * Test execution expands per mode, including reverse where supported.
* **Documentation**
  * Updated environment variable documentation for `TFT_DEFAULT_TARGET_ACCESS_MODE` and `TFT_ENABLE_TARGET_ACCESS_SUBTESTS`.

### DIFF
--- a/testSettings.py
+++ b/testSettings.py
@@ -19,6 +19,7 @@ class TestSettings:
     cfg_descr: testConfig.ConfigDescriptor
     instance_index: int
     reverse: bool
+    target_access_mode: tftbase.TargetAccessMode = tftbase.TargetAccessMode.IP
 
     event_server_alive: threading.Event = dataclasses.field(
         init=False, default_factory=threading.Event
@@ -138,7 +139,7 @@ class TestSettings:
         return self.test_case_id.info.connection_mode
 
     def get_test_info(self) -> str:
-        return f"""type={self.connection.test_type.name}, test-case={self.test_case_id.name}: {self.client_pod_type.name} pod to {self.connection_mode.name} to {self.server_pod_type.name} pod - {self.test_case_id.info.node_location}
+        return f"""type={self.connection.test_type.name}, test-case={self.test_case_id.name}: {self.client_pod_type.name} pod to {self.connection_mode.name} to {self.server_pod_type.name} pod - {self.test_case_id.info.node_location}, target-access={self.target_access_mode.name}
         Client Node: {self.node_client.name}
             Tenant={self.client_is_tenant}
             Index={self.client_index}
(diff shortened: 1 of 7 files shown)


wizhaoredhat:
@Venkat-Kunaparaju I think we should by default use the Service DNS names and not use the IP. It can catch DNS bugs which would be for sure useful.

Venkat-Kunaparaju:
> @Venkat-Kunaparaju I think we should by default use the Service DNS names and not use the IP. It can catch DNS bugs which would be for sure useful.

I'm on board for this. Will make the change.

Venkat-Kunaparaju:
Updated the default access method to use Service names, but added flag to update the default service access method if DNS is unavailable or if access via IP is still needed.

testSettings.py
```diff
@@ -151,7 +152,7 @@ def get_test_str(self) -> str:
         direction = ""
         if self.reverse:
             direction = "-REV"
-        return f"{self.test_case_id.name}-{self.client_pod_type.name}_TO_{self.connection_mode.name}_TO_{self.server_pod_type.name}-{self.test_case_id.info.node_location}{direction}"
+        return f"{self.test_case_id.name}-{self.client_pod_type.name}_TO_{self.connection_mode.name}_TO_{self.server_pod_type.name}-{self.test_case_id.info.node_location}{direction}-{self.target_access_mode.name}"
```
wizhaoredhat: Wouldn't this always add the service access mode to the test?
I would keep the test case name more mean and rely on the fact that by default we are using DNS names.

tftbase.py
```diff
@@ -538,6 +567,56 @@ class ConnectionMode(Enum):
...
+    ConnectionMode.LOAD_BALANCER,
+)
+
+
+def get_default_target_access_mode(
+    connection_mode: ConnectionMode,
+) -> TargetAccessMode:
+    override = get_tft_default_target_access_mode_override()
+    if override == TargetAccessMode.IP:
+        return TargetAccessMode.IP
```
wizhaoredhat: Looks like override is only for IP and not SERVICE_NAME as documented.
I think if an override is explicit then we should use that override; in this case SERVICE_NAME would be forced even in the cases where coredns is not running and all the test fail.

wizhaoredhat (changes requested):

testSettings.py
Venkat-Kunaparaju: Good point. For the default case I will not update the name, which is the DNS name.

Venkat-Kunaparaju:
Updated to force override and test case name.

wizhaoredhat (approved):

PR #298, opened 2026-06-03
Support IPv4, IPv6, and dual-stack flow tests

Venkat-Kunaparaju:
Add an ip_family connection option so tests can run over IPv4, IPv6, or both families. Keep IPv4 as the default to preserve existing behavior, and include the active family in test metadata, log output, result summaries, and generated test names.

Select pod, service, LoadBalancer, secondary-network, and UDN addresses by the active IP family. For dual-stack connections, run each test once per family so v4 and v6 results are reported separately.

Update HTTP and external/podman paths for IPv6 traffic, including IPv6 URL formatting, HTTP server binding, host route lookup, and podman port publishing. Mark generated Services as PreferDualStack so Kubernetes allocates dual-stack service IPs when available while still falling back to single-stack clusters.

## Release Notes

* **New Features**
  * Added support for configurable IP family preferences (IPv4, IPv6, dual-stack) per connection in test configurations
  * Tests now automatically execute across all supported IP families, enhancing test coverage
  * Services are configured with dual-stack IP family preference support

* **Documentation**
  * Updated configuration reference with new `ip_family` field options and examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,31 +53,32 @@ tft:
         instances: (8)
         reverse: "(9)"
         duration: "(10)"
+        ip_family: "(11)"
         server:
-          - name: "(11)"
-            persistent: "(12)"
-            sriov: "(13)"
-            default_network: "(14)"
-            secondary_network_nad: "(15)"
+          - name: "(12)"
+            persistent: "(13)"
+            sriov: "(14)"
+            default_network: "(15)"
+            secondary_network_nad: "(16)"
         client:
-          - name: "(16)"
-            sriov: "(17)"
-            default_network: "(18)"
-            secondary_network_nad: "(19)"
+          - name: "(17)"
+            sriov: "(18)"
+            default_network: "(19)"
+            secondary_network_nad: "(20)"
         plugins:
-          - name: (20)
-            test_cases: (21)
-          - name: (20)
-        secondary_network_nad: "(22)"
-        resource_name: "(23)"
-        cpu_request: "(24)"
-        cpu_limit: "(25)"
-        mem_request: "(26)"
-        mem_limit: "(27)"
-    privileged_pod: (28)
-    capabilities_pod: (29)
-kubeconfig: (30)
-kubeconfig_infra: (30)
+          - name: (21)
+            test_cases: (22)
+          - name: (21)
+        secondary_network_nad: "(23)"
+        resource_name: "(24)"
+        cpu_request: "(25)"
+        cpu_limit: "(26)"
+        mem_request: "(27)"
+        mem_limit: "(28)"
+    privileged_pod: (29)
+    capabilities_pod: (30)
+kubeconfig: (31)
+kubeconfig_infra: (31)
 ```
 
 1. "name" - This is the name of the test. Any string value to identify the test.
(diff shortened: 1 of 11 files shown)


Venkat-Kunaparaju:
Need to update CI lanes for v6/dual-stack support before merging.

README.md
```diff
@@ -156,37 +157,38 @@ kubeconfig_infra: (30)
...
-16. "name" - The node name of the client.
-17. "sriov" - Whether SRIOV should be used for the client pod. Takes in "true/false"
-18. "default_network" - (Optional) The name of the default network that the sriov pod would use.
-18a. "args" - (Optional) Extra command-line arguments to pass to the test tool (iperf3, simple-tcp-server-client). Supported for iperf-tcp, iperf-udp, and simple test types. Can be a string or list of strings.
-19. "secondary_network_nad" - (Optional) The secondary network NAD for the client node. Overrides the connection-level `secondary_network_nad` for the client pod. Useful when server and client require different NADs.
-20. "name" - (Optional) list of plugin names
+11. "ip_family" - (Optional) The IP address family to use for this connection. Accepted values: `ipv4` (default), `ipv6`, `dual-stack`. When omitted, the connection uses IPv4, which is the same behavior as before this field existed. On a dual-stack cluster, this lets you choose which address family (or both) to exercise per-connection.
```
wizhaoredhat, line 160: Shouldn't the default be dual-stack? It should work when the cluster is IPv4 only right? Or no..?

wizhaoredhat:
> Need to update CI lanes for v6/dual-stack support before merging.

@Venkat-Kunaparaju Was this done already?

Venkat-Kunaparaju:
> > Need to update CI lanes for v6/dual-stack support before merging.
> 
> @Venkat-Kunaparaju Was this done already?

Not yet. Need to deploy ovn-k in dpu-sim with custom helm values to enable V6. Once we have a framework for doing this in here, then I was planning to put in that update.

README.md
Venkat-Kunaparaju: With the way I implemented it now, if we specify dual-stack it will assume that v4 and v6 IPs are both present, and iterate over each IP and run the relevant test. IPv4 mode means that the IPv4 address of nodes/pods are used. Since the current tests have been running in v4 environments, I kept it as the default.

task.py
```diff
@@ -601,12 +601,23 @@ def run_oc_get(
             namespace=self._get_run_oc_namespace(namespace),
         )
 
+    def _pick_pod_ip_for_family(self, y: dict[str, Any]) -> Optional[str]:
```
wizhaoredhat, line 604: Can you add commentary for this function?

That it looks for active_ip_family and chooses the pod ip based on that?

task.py
```diff
@@ -634,7 +648,7 @@ def get_pod_ip(self) -> str:
                     nad = self._get_effective_secondary_network_nad()
                     for network in network_status:
                         if network["name"] == nad:
-                            pod_ip = network["ips"][0]
+                            pod_ip = self.ts.active_ip_family.pick_ip(network["ips"])
```
wizhaoredhat, line 651: Shouldn't this use: _pick_pod_ip_for_family ?

tftbase.py
```diff
@@ -382,6 +385,32 @@ def eval_binary_opt_in(
     return a, b
 
 
+class IpFamily(Enum):
```
wizhaoredhat, line 388: Please document this new class. What it does. What the methods do.

task.py
Venkat-Kunaparaju: `_pick_pod_ip_for_family` is for `status.podIPs`. Since this is for 2ndary networks, the annotation for the ip to use exists in `network["ips"]`.

testSettings.py
```diff
@@ -19,6 +19,7 @@ class TestSettings:
     cfg_descr: testConfig.ConfigDescriptor
     instance_index: int
     reverse: bool
+    active_ip_family: tftbase.IpFamily = tftbase.IpFamily.IPV4
```
wizhaoredhat, line 22: Please document this class's members. Like active_ip_family what does it do. Why is it called "active". Why we can't use the input parameter ip_family directly?

"active_ip_family" makes it seem like the ip family can change when it should be static.

Venkat-Kunaparaju:
Still need to address comments and rebase. Will pick this back up eventually since it is important, and also need to make sure dpu-sim can also support v4/v6/dual-stack testing so that we can properly invoke this in CI. 

For now this is on the backburner.

PR #302, opened 2026-06-21
NVIDIA-554 Add EgressIP support (2)

josecastillolema:
Based on #247, reimplemented with source IP verification.

- Add `POD_TO_EXTERNAL_EGRESS` test case type (68) for EgressIP testing 
- Add `egress_ip` connection parameter with IP/subnet validation against `k8s.ovn.org/node-primary-ifaddr`
- Add `create_egressip()` on ClientTask: labels node, renders and applies EgressIP CR, polls `.status.items` until assignment confirmed (120s timeout)
- Capture iperf3 server stdout and verify source IP matches the configured EgressIP after each test run
- Add EgressIP CR and node label cleanup in `_cleanup_previous_testspace()`

### Source IP verification

The key addition over #247 is end-to-end verification: after iperf completes, the server's JSON output is parsed for `start.connected[].remote_host` and compared against the configured EgressIP. The test fails if the source IP doesn't match, catching cases where OVN doesn't program the NAT rules and traffic falls back to default SNAT.

### Usage
```yaml
connections:
  - name: "test"
    type: iperf-tcp
    egress_ip:
      ip: "192.168.1.100"
      node: "worker-1"  # optional, defaults to client node
    external_server_ip: "10.0.0.1"  # optional
```

## Test plan

- [x] Unit tests pass (pytest — 47 tests)
- [x] Pre-commit hooks pass (black, flake8, mypy)
- [x] Verified with live EgressIP test on a couple clusters (9.33 Gbps, source IP verified)

Assisted-by: Claude 🤖

## Summary by CodeRabbit

* **New Features**
  * Added a new pod-to-external egress iPerf test case (`POD_TO_EXTERNAL_EGRESS`) with configurable TCP/UDP thresholds.
  * Introduced optional EgressIP-based routing with automatic EgressIP provisioning and assignment verification (including support for externally targeted server IP selection).
* **Bug Fixes**
  * Improved verification by extracting the server’s remote host/source IP and failing the run when it doesn’t match the expected EgressIP.
  * Enhanced target IP discovery with more resilient routing/lookup behavior.
* **Chores / Tests**
  * Updated evaluation config outputs, added unit tests for EgressIP parsing and server output handling, and expanded cleanup to remove EgressIP-related state.

### DIFF
--- a/task.py
+++ b/task.py
@@ -1502,7 +1502,7 @@ class ClientTask(Task, ABC):
                 logger.warning(f"Could not validate EgressIP subnet: {e}")
 
         self.run_oc(
-            f"label node {egress_node} k8s.ovn.org/egress-assignable=true --overwrite",
+            f"label node {egress_node} k8s.ovn.org/egress-assignable=true tft-tests-egress-node=true --overwrite",
             namespace=None,
             die_on_error=True,
         )
@@ -1653,7 +1653,9 @@ class ClientTask(Task, ABC):
                         msg="EgressIP verification failed: could not parse "
                         "server output for source IP",
                     )
-                elif actual_ip != expected_ip:
+                elif ipaddress.ip_address(actual_ip) != ipaddress.ip_address(
+                    expected_ip
+                ):
                     self._result = dataclasses.replace(
                         self._result,
                         success=False,
--- a/task.py
+++ b/task.py
@@ -1576,13 +1576,6 @@ class ClientTask(Task, ABC):
             logger.debug(f"get_target_ip() NodePortIP connection to {ip}")
             return ip
         elif self.connection_mode == ConnectionMode.EXTERNAL_IP:
-            ext_ip = self.ts.connection.external_server_ip
-            if ext_ip is not None:
-                logger.debug(
-                    f"get_target_ip() External connection to configured "
-                    f"server {ext_ip}"
-                )
-                return ext_ip
             host_ip = self.get_host_ip()
             logger.debug(f"get_target_ip() External connection to host {host_ip}")
             return host_ip
@@ -1614,9 +1607,23 @@ class ClientTask(Task, ABC):
         return self.port
 
     def get_host_ip(self) -> str:
-        """Get the host's IP address for EXTERNAL_IP mode."""
-        # Get the IP from default route using JSON output
-        r = self.lh.run("ip -j route get 1")
+        """Get the host's IP address for EXTERNAL_IP mode.
+
+        Probes the route toward the server node so that the kernel picks
+        the correct source interface (e.g. a VPN tunnel rather than the
+        default LAN gateway).
+        """
+        server_name = self.ts.node_server.name
+        r = self.lh.run(f"getent hosts {shlex.quote(server_name)}")
+        if r.success and r.out.strip():
+            node_ip = r.out.split()[0]
+        else:
+            node_ip = "1"
+            logger.debug(
+                f"get_host_ip() could not resolve {server_name}, "
+                "falling back to default route"
+            )
+        r = self.lh.run(f"ip -j route get {node_ip}")
         if r.success and r.out.strip():
             try:
                 routes = json.loads(r.out.strip())
@@ -1626,7 +1633,7 @@ class ClientTask(Task, ABC):
                     return ip
             except (json.JSONDecodeError, KeyError, IndexError):
                 pass
-        raise RuntimeError("Failed to get host IP address from default route")
+        raise RuntimeError("Failed to get host IP address from route lookup")
 
     def aggregate_output(self, tft_result_builder: tftbase.TftResultBuilder) -> None:
         if (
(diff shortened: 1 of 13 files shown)


task.py
```diff
@@ -1525,6 +1628,45 @@ def get_host_ip(self) -> str:
...
+                        msg="EgressIP verification failed: could not parse "
+                        "server output for source IP",
+                    )
+                elif actual_ip != expected_ip:
+                    self._result = dataclasses.replace(
+                        self._result,
+                        success=False,
+                        msg=f"EgressIP verification failed: expected source "
+                        f"IP {expected_ip} but got {actual_ip}",
+                    )
```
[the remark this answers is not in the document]
josecastillolema: Fixed, check again

trafficFlowTests.py
```diff
@@ -335,6 +335,25 @@ def _cleanup_previous_testspace(
...
+                namespace=None,
+                may_fail=True,
+            )
+            if r.success and r.out.strip():
+                for node_name in r.out.strip().split():
+                    client.oc(
+                        f"label node {node_name} k8s.ovn.org/egress-assignable-",
+                        namespace=None,
+                        may_fail=True,
+                    )
```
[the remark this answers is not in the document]
josecastillolema: Fixed, check again

task.py
```diff
@@ -1397,9 +1410,16 @@ def _run_cmd(cmd: str) -> BaseOutput:
             return BaseOutput.from_cmd(res, success=True if may_fail else None)
 
         def _thread_action() -> BaseOutput:
-            return _run_cmd(cmd)
+            return _run_cmd(cmd, capture_server_output=True)
 
         def _cancel_action() -> None:
+            if self.server_stdout is None:
```
tsorya, line 1416: We want tp take logs before cancelling ?  Sorry didn't fully understand the change

josecastillolema: For `EXTERNAL_IP` mode, the server runs as a podman container with `--one-off`. When the test finishes, `_cancel_action` kills the container via `_run_cmd(cancel_cmd)`. If _thread_action didn't capture the output (e.g. the container already exited before we could read stdout), we need to grab it via podman logs before the cancel destroys the container. This is the only chance to get the server's JSON output for EgressIP source IP verification.

Venkat-Kunaparaju:
Is it possible to enable the test in one of the ci lanes to see how it runs? Probably be best to do it in `full-suite`.

task.py
Venkat-Kunaparaju: If we set the server to `persistent: true`, will it still be able to grab the logs in this case?

task.py
```diff
@@ -1479,7 +1576,13 @@ def get_target_ip(self) -> str:
             logger.debug(f"get_target_ip() NodePortIP connection to {ip}")
             return ip
         elif self.connection_mode == ConnectionMode.EXTERNAL_IP:
-            # For port forwarding, connect to the host's IP, not the container's internal IP
+            ext_ip = self.ts.connection.external_server_ip
```
Venkat-Kunaparaju: If we do specify an `external_server_ip`, will it also still create the podman server? If it does then the podman server won't be used since we are using this specified IP for the connection instead. Maybe we should have a way to not deploy the podman server to conserve resources.

task.py
josecastillolema: Yes, it should still work. The persistent flag has no effect on the podman/EXTERNAL_IP path.

josecastillolema:
> Is it possible to enable the test in one of the ci lanes to see how it runs? Probably be best to do it in `full-suite`.

podman isn't currently used in any CI lane. I would need to check if it's available and we'd also need a valid egress IP from the KIND node subnet. Correct me if I am wrong but both depend on the dpu-sim cluster setup, I need to verify that first before enabling it in full-suite.

task.py
josecastillolema: Your comment get me thinking and I have deleted the `external_server_ip` option. It does not make sense because we cannot verify the source address when using external servers.

Venkat-Kunaparaju:
> > Is it possible to enable the test in one of the ci lanes to see how it runs? Probably be best to do it in `full-suite`.
> 
> podman isn't currently used in any CI lane. I would need to check if it's available and we'd also need a valid egress IP from the KIND node subnet. Correct me if I am wrong but both depend on the dpu-sim cluster setup, I need to verify that first before enabling it in full-suite.

Yeah agreed, I wasnt sure as well probably needs more looking into. It does depend on the dpu-sim setup, I am currently working on making the ci lanes with the dpu-sim more robust.

Venkat-Kunaparaju (approved):
LGTM, thanks for adding this

wizhaoredhat:
@josecastillolema I think you added a new API to the yaml file. Could you create a new PR with changes to the README?

josecastillolema:
> @josecastillolema I think you added a new API to the yaml file. Could you create a new PR with changes to the README?

Sure @wizhaoredhat https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/309

PR #304, opened 2026-06-25
Delete CUDN after ns and UDN deletion

Venkat-Kunaparaju:
Cleanup of stale udn resources gets stalled when executing CUDN cleanup first. 

To prevent this, CUDN cleanup is moved to the last step after ns and UDN deletion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of user-defined network resources during test runs.
  * Ensured stale network namespaces and associated resources are removed consistently.
  * Added waiting for namespace removal to complete before proceeding.

### DIFF
--- a/trafficFlowTests.py
+++ b/trafficFlowTests.py
@@ -285,20 +285,34 @@ class TrafficFlowTests:
         tft = cfg_descr.get_tft()
         udn_ns = tftbase.get_udn_namespace(tft.namespace)
         client = cfg_descr.tc.client_tenant
-        client.oc(
-            "delete clusteruserdefinednetwork -l tft-tests",
-            namespace=None,
-            may_fail=True,
-        )
         if client.oc_get(f"namespace/{udn_ns}", may_fail=True) is None:
+            client.oc(
+                "delete clusteruserdefinednetwork -l tft-tests --timeout=120s",
+                namespace=None,
+                may_fail=True,
+            )
             return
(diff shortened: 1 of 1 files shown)


trafficFlowTests.py
```diff
@@ -285,20 +285,34 @@ def _cleanup_stale_udn(self, cfg_descr: ConfigDescriptor) -> None:
...
+                namespace=None,
+                may_fail=True,
+            )
             return
         logger.info(
             f"Found existing UDN namespace {udn_ns} from a previous run, cleaning up"
         )
-        client.oc(f"delete userdefinednetwork -l tft-tests -n {udn_ns}", may_fail=True)
-        client.oc(f"delete namespace {udn_ns}", may_fail=True)
+        client.oc(
```
wizhaoredhat: It would appear that _cleanup_udn is doing something different than here (i.e. CUDN is deleted first). Shouldn't we consolidate the "cleanup" logic for both stale udns and while we cleanup? It should be at the very least consistent (if maintaining that consistency is easy to miss then we probably need a function).

wizhaoredhat (changes requested):

Venkat-Kunaparaju:
Updated to consolidate the regular cleanup and the stale cleanup logic for UDNs. So they both exercise this change.

wizhaoredhat (approved):

PR #305, opened 2026-06-25
Skip namespace setup for UDN-only test runs

Venkat-Kunaparaju:
The namespace used for non-UDN tests is still created when the test suite is only UDN tests, which operate in a separate namespace. 

Update so that the default namespace is not created when the test suite is UDN only.

### DIFF
--- a/trafficFlowTests.py
+++ b/trafficFlowTests.py
@@ -31,6 +31,9 @@ class TrafficFlowTests:
         self._udn_ns_created: bool = False
         self._udn_setup_done: bool = False
 
+    def _has_non_udn_tests(self, cfg_descr: ConfigDescriptor) -> bool:
+        return any(not tc.is_udn for tc in cfg_descr.get_tft().test_cases)
+
     def _configure_namespace(
         self, cfg_descr: ConfigDescriptor, *, namespace: str | None = None
     ) -> bool:
@@ -166,6 +169,8 @@ class TrafficFlowTests:
 
     def _setup_secondary_nad(self, cfg_descr: ConfigDescriptor) -> None:
         tft = cfg_descr.get_tft()
+        if not self._has_non_udn_tests(cfg_descr):
+            return
         if not any(
             not tc.is_udn
             and tc.info.connection_mode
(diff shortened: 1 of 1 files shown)


wizhaoredhat (changes requested):
In the cleanup (with namespace) scenarios, I believe you haven't touched to make it UDN aware may expect the namespace to be there. E.g. when deleting pods from a specific namespace.

Therefore the tests will fail when the namespace doesn't exist during cleanup.

But this whole logic to exclude creating namespaces... I question if it is even worth doing? Because now you have to worry about whether a namespace exists or not. It doesn't look fun when you think about the creation and cleanup parts.

Venkat-Kunaparaju:
> In the cleanup (with namespace) scenarios, I believe you haven't touched to make it UDN aware may expect the namespace to be there. E.g. when deleting pods from a specific namespace.
> 
> Therefore the tests will fail when the namespace doesn't exist during cleanup.
> 
> But this whole logic to exclude creating namespaces... I question if it is even worth doing? Because now you have to worry about whether a namespace exists or not. It doesn't look fun when you think about the creation and cleanup parts.

This cleanup is a no-op when the namespace does not exist. There is a selector base cleanup, `kubectl delete ... -l`, commands return without failing when no resources are found, and namespace deletion is already gated by `ns_created`, which only deletes the namespace if it was created.

wizhaoredhat:
> > In the cleanup (with namespace) scenarios, I believe you haven't touched to make it UDN aware may expect the namespace to be there. E.g. when deleting pods from a specific namespace.
> > Therefore the tests will fail when the namespace doesn't exist during cleanup.
> > But this whole logic to exclude creating namespaces... I question if it is even worth doing? Because now you have to worry about whether a namespace exists or not. It doesn't look fun when you think about the creation and cleanup parts.
> 
> This cleanup is a no-op when the namespace does not exist. There is a selector base cleanup, `kubectl delete ... -l`, commands return without failing when no resources are found, and namespace deletion is already gated by `ns_created`, which only deletes the namespace if it was created.

Ah I see. Ok, so we do a blanket cleanup of everything. I was under the wrong impression that we failed on deletion if non-existent. But we selective for creation, I think this is ok to not pollute the namespaces.

wizhaoredhat (approved):

PR #307, opened 2026-06-29
Expand UDN/CUDN testing and add primary no-overlay support

Venkat-Kunaparaju:
This PR rebuilds UDN coverage around configurable primary networks and explicit secondary network permutations.

- Add test-level primary network configuration for UDN/CUDN mode, Layer2/Layer3 topology, and overlay/no-overlay transport while preserving the existing UDN Layer3 overlay defaults.
- Replace the legacy secondary and localnet tests with same-node and different-node pod-to-pod cases for UDN/CUDN Layer3, UDN/CUDN Layer2, and CUDN localnet.
- Retire test IDs 48-59 and introduce the replacement secondary tests as IDs 69-78.
- Support attaching multiple NADs to shared pre-provisioned pods and select the exact NAD and IP required by each test without changing regular secondary NAD behavior.
- Consolidate network rendering into shared UDN and CUDN templates with dedicated names and CIDRs for each secondary permutation.
- Create and clean up RouteAdvertisements for unmanaged primary Layer3 no-overlay CUDNs using the configured FRRConfiguration selector.

The legacy secondary service tests are removed because their Services use the primary network and do not provide meaningful coverage of the secondary interface.

## Testing

Used KIND cluster in both overlay and no-overlay mode for validation

- Ran the positive UDN/CUDN configs verified that the framework rendered the expected UDN, CUDN, NAD, and pod resources from `config.yaml`.
- Verified primary network rendering for UDN and CUDN Layer3/Layer2 permutations, including no-overlay CUDN generation. The generated CUDNs matched the requested config, including `transport: NoOverlay`, `routing: Managed` or `Unmanaged`, `outboundSNAT: Enabled`, and the expected primary CIDR/hostSubnet values.
- Verified unmanaged no-overlay CUDNs create a matching `RouteAdvertisements` object using the configured `frr_configuration_selector`.
- Verified the secondary network matrix renders distinct networks and CIDRs for each permutation, including CUDN Layer3, UDN Layer3, CUDN Layer2, UDN Layer2, and CUDN localnet.
- Ran coverage with both `pre_provision: false` and `pre_provision: true`. In both modes, pods were created successfully and attached to the expected primary and secondary interfaces.
- In `pre_provision: true` mode, verified shared pods received all required NAD attachments and that each test selected the correct NAD/IP for its specific network.
- Confirmed cleanup removed the generated UDN/CUDN, `RouteAdvertisements`, NADs, pods, services, and namespaces created by the tests.

## Summary by CodeRabbit

* **New Features**
  * Added `udn_primary_network` to configure UDN/CUDN `mode`, `topology`, `transport` and optional FRR route advertisement selection.
  * Expanded UDN testing coverage for CUDN/UDN Layer2/Layer3 and Localnet, including support for no-overlay routing.

* **Bug Fixes**
  * Corrected how secondary network NADs are selected for pre-provisioned shared pods and adjusted UDN pod IP/secondary IP behavior to match the active network spec.

* **Documentation**
  * Updated README connection templates, environment variables, and test-case numbering/grouping; clarified `secondary_network_nad` behavior for tests 70–79.

* **Tests**
  * Refreshed generated eval configs, thresholds, and unit tests for the new UDN/CUDN scenarios and IDs.


tftbase.py
```diff
@@ -469,20 +541,14 @@ class TestCaseType(Enum):
...
-    HOST_TO_LOAD_BALANCER_TO_HOST_SAME_NODE = 66
-    HOST_TO_LOAD_BALANCER_TO_HOST_DIFF_NODE = 67
+    POD_TO_LOAD_BALANCER_TO_POD_SAME_NODE = 54
+    POD_TO_LOAD_BALANCER_TO_POD_DIFF_NODE = 55
+    POD_TO_LOAD_BALANCER_TO_HOST_SAME_NODE = 56
+    POD_TO_LOAD_BALANCER_TO_HOST_DIFF_NODE = 57
+    HOST_TO_LOAD_BALANCER_TO_POD_SAME_NODE = 58
+    HOST_TO_LOAD_BALANCER_TO_POD_DIFF_NODE = 59
+    HOST_TO_LOAD_BALANCER_TO_HOST_SAME_NODE = 60
+    HOST_TO_LOAD_BALANCER_TO_HOST_DIFF_NODE = 61
```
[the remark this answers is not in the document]
Venkat-Kunaparaju: Silently remapping is okay, the old localnet tests were unused.

trafficFlowTests.py
```diff
@@ -54,13 +54,169 @@ def _configure_namespace(
...
+                        and condition.get("type") == "TransportAccepted"
+                    ):
+                        if condition.get("status") == "True":
+                            return
+                        break
+            _time.sleep(2)
+        logger.info(
+            f"ClusterUserDefinedNetwork/{cudn_name} TransportAccepted not True, continuing"
+        )
+
```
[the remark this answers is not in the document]
Venkat-Kunaparaju: This is correct, it should result in test failure if TransportAccepted never becomes True. It will allow the rest of the tests to continue, without this test being a blocking case, and will show clearly when TransportAccepted is not true when the test fails, allowing the user to see what went wrong.

testConfig.py
```diff
@@ -94,6 +97,24 @@ def _construct_capabilities_pod(
...
+    for key, value in arg.items():
+        if not key:
+            raise pctx.value_error("label key cannot be empty")
+        if value is None:
+            result[key] = ""
+        elif isinstance(value, str):
+            result[key] = value
+        else:
+            raise pctx.value_error(
+                f"expects string label values but got {repr(value)}",
```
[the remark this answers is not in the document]
Venkat-Kunaparaju: Not fixing, this same violation is seen in other parts of the repo. Maintaining consistency with it for now.

README.md
```diff
@@ -76,8 +76,21 @@ tft:
         mem_limit: "(27)"
     privileged_pod: (28)
     capabilities_pod: (29)
-kubeconfig: (30)
-kubeconfig_infra: (30)
+    udn_primary_network: # (30)
```
wizhaoredhat: The disadvantage of this approach is that you can only test 1 set of UDN primary and secondary setups either the default values or modified values.
There is no way to, for example, test no-overlay and overlay with the same YAML file.

I am not sure what the best way to fix such an issue. The previous approach can increase the test case numbers exponentially which is not sustainable.

We can change the yaml file scheme to have a list of UDN tests that can be customized as you have it right now. But the disadvantage is when trying to compare notes on what people test, using test case numbers is universal (for non UDN test cases) but for UDN you must send each other the YAML file config.

Venkat-Kunaparaju: To your first point, we can only have 1 primary UDN. So we can't concurrently test no-overlay and overlay, or layer3 and layer2 primary. So it will always be an either or. We can concurrently have multiple secondary networks, but the approach before having different test cases for different secondary networks leads to too many test cases.

I agree that it can lead to confusion in comparing test outputs. And I don't want to add test cases for every single combination like you mentioned. 

What I can do is add logic in the test cases output to specify the configuration of the UDNs, so it is more clear. Similar to how we specify the connection type in the test case output.

wizhaoredhat: "So we can't concurrently test no-overlay and overlay, or layer3 and layer2 primary"
Not concurrently but different isolated tests would be ok to switch between these right?

Venkat-Kunaparaju: Yeah if we have separate `tft` entries it is possible. UDNs are currently created and deleted not per test case but per `tft` entry. 

So it is possible to test overlay and no-overlay in the same YAML file if it has multiple `tft` entries with different parameters.

wizhaoredhat: Why are they not cleaned up and recreated per test case?

Venkat-Kunaparaju: Regardless of `pre_provision` enabled/disabled, it will do it per `tft`. Pretty sure this was done in the initial PR @aserdean put in for UDN support.

Also the UDNs are created in a namespace separate from the CDN tests, so the UDNs being there for the whole time are only relevant for UDN specific pods and tests.

wizhaoredhat: Could you think of anyway to keep the configuration more lean. I feel like this adds too much configuration options that changes the behaviour of the test quite significantly.

Maybe we can have a balance such as "mode" = "udn" vs. "cudn" as seperate test cases?

Defiantly when printing results out, we need to differentiate the test by the additional config. Otherwise when discussing results between teams or even companies can get very confusing and possibly have people go down the wrong path.

Venkat-Kunaparaju: New framework:

The implementation for configuring the primary UDN tests will stay mostly the same. Only 1 set of `UDN_PRIMARY ` will be there, and configuration of the UDN will be done through config.yaml. 

Secondary UDN tests will not be configured via config.yaml, but rather have test cases for each permutation supported. CUDN/UDN (2) x Layer3/Layer2/Localnet (3) = 6 tests, but UDN x Localnet is not supported so **5** sets of tests. The service tests that we currently have for Secondary/Localnet UDN tests will use the primary network, so there is not enough value in keeping them now when we have to permutate this 5 times. 

So we will now have only the follow secondary UDN tests:

```
CUDN_LAYER3_POD_TO_POD_SAME_NODE
CUDN_LAYER3_POD_TO_POD_DIFF_NODE
UDN_LAYER3_POD_TO_POD_SAME_NODE
UDN_LAYER3_POD_TO_POD_DIFF_NODE

CUDN_LAYER2_POD_TO_POD_SAME_NODE
CUDN_LAYER2_POD_TO_POD_DIFF_NODE
UDN_LAYER2_POD_TO_POD_SAME_NODE
UDN_LAYER2_POD_TO_POD_DIFF_NODE

CUDN_LOCALNET_POD_TO_POD_SAME_NODE
CUDN_LOCALNET_POD_TO_POD_DIFF_NODE
```

Venkat-Kunaparaju:
Made the changes based on the discussion. Also updated PR title and description with up to date information.

Venkat-Kunaparaju:
https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/313

Used this PR to run dpu-sim ci lanes that run in overlay and no-overlay mode, running these new UDN tests. 

Custom dpu-sim branch: https://github.com/Venkat-Kunaparaju/dpu-simulator/tree/udn-test

Pod->external in no-overlay broken until https://github.com/ovn-kubernetes/ovn-kubernetes/pull/6620 goes through.

manifests/secondary-nad.yaml.j2
```diff
@@ -13,8 +15,7 @@ spec:
     "Role": "secondary",
     "topology": {{ topology }},
     "subnets": {{ subnets }},
-    "mtu": {{ mtu }},{% if has_resource_name %}
-    "resourceName": {{ resource_name }},{% endif %}
```
wizhaoredhat, line 17: Why was resourceName removed from the CNI config?

task.py
```diff
@@ -447,6 +447,22 @@ def _get_pod_secondary_network_nads(self) -> tuple[str, ...]:
             return (self._get_effective_secondary_network_nad(),)
         return ()
 
+    def _get_sriov_resource_quantity(
```
wizhaoredhat, line 469: Let's add a function header and explain how we calculate resource quantity here. It seems quite complicated so far.

README.md
```diff
@@ -235,6 +240,18 @@ The primary CIDR defaults to `15.1.0.0/16`. Secondary CIDRs default to `15.2.0.0
 
 `udn_primary_network` supports `mode` values `udn` and `cudn`, `topology` values `layer3` and `layer2`, and `transport` values `overlay` and `no-overlay`. `no-overlay` requires `mode: cudn` and `topology: layer3`.
 
+For unmanaged no-overlay primary CUDNs, set `frr_configuration_selector` to create a RouteAdvertisements object that advertises the selected CUDN pod network. The selector is a map of `frrConfigurationSelector.matchLabels` labels. When `TFT_UDN_NO_OVERLAY_ROUTING_MANAGED` is true, RouteAdvertisements are not created.
```
wizhaoredhat: Why is TFT_UDN_NO_OVERLAY_ROUTING_MANAGED needed when you can have an empty frr_configuration_selector?

task.py
```diff
@@ -404,11 +404,20 @@ def _get_template_args_test_image(self) -> str:
         """Get the test image, auto-selecting RDMA image for ib-* test types."""
         return tftbase.get_tft_test_image_for_type(self.ts.connection.test_type)
 
+    def _shares_pre_provisioned_secondary_pod(self) -> bool:
```
wizhaoredhat, line 408: Changes to task.py is quite confusing to me. It can probably help with some commentary to explain the new functionality of these functions.

manifests/secondary-nad.yaml.j2
Venkat-Kunaparaju: The annotation was in the wrong spot before the for this nad manifest. I updated it to align with what is used in UDNs, so it gets picked up properly.

README.md
Venkat-Kunaparaju: When we have an empty `frr_configuration_selector ` it could also mean that the user has there own RA that they setup to carry TFT traffic. For example, there might already be some default RA that matches on all pods, not just TFT pods, which can pick it up. So `frr_configuration_selector` is just a feature to create that RA for you or rely on some existing RA that the user has already setup.

task.py
Venkat-Kunaparaju: I will put in an update to document my changes more.

wizhaoredhat:
@Venkat-Kunaparaju Could you rebase the conflicts so that I can review?

wizhaoredhat (approved):
@Venkat-Kunaparaju
 Can you rebase, so it uses the latest CI changes? Then once CI is green we can merge

Venkat-Kunaparaju:
> @Venkat-Kunaparaju Can you rebase, so it uses the latest CI changes? Then once CI is green we can merge

Will do, I also need to rebase my downstream changes for dpu-simulator, and then run these tests that this PR is adding against that again so that we can make sure these tests are all green to.

wizhaoredhat:
@coderabbitai resume

PR #308, opened 2026-07-03
Add NetworkPolicy namespaceSelector test for host-network traffic

josecastillolema:
Adds test case 69 (`HOST_TO_POD_NP_NS_SELECTOR_ALLOW`) that verifies NetworkPolicy namespaceSelector matching against the OVN-K host-network namespace. A hostNetwork client sends traffic to a normal server pod protected by a deny-all + allow-from-host-network-namespace NP.

Passes on standard OCP clusters (`openshift-host-network` populated), fails on DPF/DPU clusters where the namespace is missing.

Configurable via `TFT_HOST_NETWORK_NAMESPACE` env var (defaults to `openshift-host-network`).

Assisted-by: Claude 🤖

## Summary by CodeRabbit

* **New Features**
  * Added a new traffic test mode that applies NetworkPolicy ingress using a namespace selector (“allow”).
  * Introduced a new host-to-pod eval test case (ID 69) with updated TCP/UDP threshold values in generated configurations.
* **Bug Fixes**
  * Improved reset/cleanup behavior so the new mode is handled consistently.
* **Tests**
  * Updated enum expectations and evaluation config outputs to include the new test case.
* **Documentation**
  * Documented the new test case ID, expanded connection-level egress IP guidance, and added/clarified the host network namespace environment variable.


josecastillolema:
cc @aserdean

Venkat-Kunaparaju:
Also update the README here with the new test case changes and the new env variable

manifests/network-policy-ns-selector.yaml.j2
```diff
@@ -0,0 +1,21 @@
...
+  policyTypes:
+    - Ingress
+  ingress:
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: {{ np_ns_name }}
+      ports:
+        - port: {{ np_ns_port }}
+          protocol: TCP
```
Venkat-Kunaparaju: Need to support UDP as well

Venkat-Kunaparaju: Follow up, maybe we can use the existing `network-policy.yaml.j2` file to render the NetworkPolicy for this test? We can add jinja logic to control when to exclude/include certain sections of the NetworkPolicy. This way we can share 1 file for the all the NetworkPolicy cases, and changes to NetworkPolicy don't have to sync across multiple files.

josecastillolema:
> Also update the README here with the new test case changes and the new env variable

Sure! Could we please get this document update PR https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/309 reviewed/merged first?

manifests/network-policy-ns-selector.yaml.j2
josecastillolema: Both comments addressed on https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/308/commits/7268b7f7a7c3af40269f87c10b6cc5396580af54

josecastillolema:
> Also update the README here with the new test case changes and the new env variable

Documentation updated.

README.md
```diff
@@ -178,11 +179,11 @@ kubeconfig_infra: (32)
 18a. "args" - (Optional) Extra command-line arguments to pass to the test tool (iperf3, simple-tcp-server-client). Supported for iperf-tcp, iperf-udp, and simple test types. Can be a string or list of strings.
 19. "secondary_network_nad" - (Optional) The secondary network NAD for the client node. Overrides the connection-level `secondary_network_nad` for the client pod. Useful when server and client require different NADs.
 20. "name" - (Optional) list of plugin names
-    | Name             | Description          |
-    | ---------------- | -------------------- |
-    | measure_cpu      | Measure CPU Usage    |
-    | measure_power    | Measure Power Usage  |
-    | validate_offload | Verify OvS Offload   |
+    | Name             | Description         |
```
Venkat-Kunaparaju: Shouldnt need to update this for this PR

README.md
```diff
@@ -320,11 +321,11 @@ traffic. Pass delegates the decision to NetworkPolicies in the namespace.
 
 Three test cases validate ANP behavior, each with the action baked into the test case type:
 
-| ID | Test Case | ANP Action | Expected Result |
-| -- | --------- | ---------- | --------------- |
-| 32 | `POD_TO_POD_ANP_ALLOW` | Allow | Traffic flows |
-| 33 | `POD_TO_POD_ANP_DENY` | Deny | Traffic blocked |
-| 34 | `POD_TO_POD_ANP_PASS_NP_DENY` | Pass (delegates to NP Deny) | Traffic blocked |
+| ID  | Test Case                     | ANP Action                  | Expected Result |
```
Venkat-Kunaparaju: Same here, I don't think its relevant for this PR

README.md
```diff
@@ -365,6 +366,33 @@ tft:
           - name: "worker-2"
 ```
 
+## NetworkPolicy namespaceSelector Tests
```
Venkat-Kunaparaju: Since this is technically still under NP set of tests and we are only adding a test case not changing config.yaml, I think its best if we either shorten this section or remove it. If we do this for every new test case the Readme will grow to be too large. WDYT?

Venkat-Kunaparaju:
Much cleaner implementation, thanks for updating. If you can add 1 more commit to add this new test case to `.github/tft-configs/diverse-subset.yaml` and then re-push and then we should see this test case run using the dpu-simulator.

Otherwise lgtm!

Venkat-Kunaparaju:
> Much cleaner implementation, thanks for updating. If you can add 1 more commit to add this new test case to `.github/tft-configs/diverse-subset.yaml` and then re-push and then we should see this test case run using the dpu-simulator.
> 
> Otherwise lgtm!

Also need to set TFT_HOST_NETWORK_NAMESPACE=ovn-host-network in `kind-tft.yml` most likely

README.md
josecastillolema: Removed this changes

README.md
josecastillolema: Removed this changes

README.md
josecastillolema: Deleted the section

josecastillolema:
> Also need to set TFT_HOST_NETWORK_NAMESPACE=ovn-host-network in `kind-tft.yml` most likely

Made ovn-host-network the default for upstream alignment within this repo.

josecastillolema:
> Much cleaner implementation, thanks for updating. If you can add 1 more commit to add this new test case to `.github/tft-configs/diverse-subset.yaml` and then re-push and then we should see this test case run using the dpu-simulator.

Done! cc @Venkat-Kunaparaju 
```
INFO   : Test ID: (69) HOST_TO_POD_NP_NS_SELECTOR_ALLOW, Test Type: IPERF_TCP, Reverse: false, TX Bitrate: 8.7028 Gbps, RX Bitrate: 2.8921 Gbps, succeeded
```

Venkat-Kunaparaju:
LGTM

tftbase.py
```diff
@@ -1482,6 +1493,13 @@ def get_client_pod_type(self, pod_type: PodType) -> PodType:
             is_server_hostbacked=False,
             is_client_hostbacked=False,
         ),
+        TestCaseTypInfo(
```
wizhaoredhat, line 1496: @josecastillolema / @Venkat-Kunaparaju 

Why is this test only for different nodes and not on the same node? If it is only for diff node, I feel like when it is HOST_TO_POD, the convention is to add DIFF_NODE or SAME_NODE to the test case name.

Venkat-Kunaparaju: Other NP tests don't use the diff_node same_node convention because all of them are applied to diff node. Which is why I don't think it is needed.

If we were to make this change it would need to be applied to other NP tests to follow convention.

PR #310, opened 2026-07-06
Add POD_TO_MGMT_PORT test case to verify pod-to-worker ovn-k8s-mp0 reachability

josecastillolema:
Adds test case 69 (`POD_TO_MGMT_PORT`) with a new ping connection type, verifying that a pod on one node can reach another node's ovn-k8s-mp0 management-port IP, catching a control-plane-to-worker mp0 unreachability issue we've observed several times in DPF clusters.

How it works:
- Client is a normal pod scheduled on the client: node
- No server pod is created (new `ConnectionMode.MGMT_PORT_IP` short-circuits server provisioning, like `EXTERNAL_IP` does).
- Target IP is derived from the server: node's `k8s.ovn.org/node-subnets` annotation (.2 address = mp0 convention).
- Client runs `ping -c 5 -W 2 <mp0-ip>`; pass/fail is exit code, no throughput involved.

Verified: black/flake8/mypy clean, 48/48 unit tests pass, and live end-to-end run against a real cluster (control-plane client → worker mp0) succeeded.

Example config file:
```yaml
tft:
  - name: "Mgmt Port Reachability Test"
    namespace: "tft-mgmt-port-test"
    test_cases: "69"
    duration: "5"
    connections:
      - name: "Connection_1"
        type: "ping"
        instances: 1
        server:
          - name: "worker1"
        client:
          - name: "master1"
kubeconfig: "~/.kube/config"
```

Assisted-by: Claude 🤖

## Summary by CodeRabbit

* **New Features**
  * Added a `ping_mgmt_port` plugin that verifies management-port reachability by pinging the server node from a tools pod.
  * Supports enabling the plugin globally or restricting it to specific test cases.
  * Automatically derives the target management IP from the server node’s subnet annotation.

* **Documentation**
  * Documented the `ping_mgmt_port` plugin, including configuration guidance and YAML examples.

* **Tests**
  * Added coverage for correct management IP derivation and validation of malformed/invalid annotation inputs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,26 @@ Test cases 37-59 run traffic over OVN-Kubernetes User Defined Networks. The fram
 
 CIDRs default to `15.1.0.0/16` (primary), `15.2.0.0/16` (secondary), and `15.3.0.0/24` (localnet), overridable via `TFT_UDN_PRIMARY_CIDR`, `TFT_UDN_SECONDARY_CIDR`, and `TFT_UDN_LOCALNET_CIDR`. The localnet physical network name defaults to `physnet`, overridable via `TFT_UDN_LOCALNET_PHYSICAL_NETWORK`. Reference manifests are in `manifests/udn-primary.yaml.j2`, `manifests/udn-secondary.yaml.j2`, and `manifests/udn-localnet.yaml.j2`.
 
+## Management Port Reachability Plugin
+
+The `ping_mgmt_port` plugin checks that the client node can reach the `ovn-k8s-mp0`
+management port interface of the server node. The target IP is derived as the `.2`
+address of the server node's `k8s.ovn.org/node-subnets` annotation. Add it to a
+connection's plugin list:
+
+```yaml
+plugins:
+  - ping_mgmt_port
+```
+
+or scoped to specific test cases:
+
+```yaml
+plugins:
+  - name: ping_mgmt_port
+    test_cases: [POD_TO_POD_DIFF_NODE]
+```
+
 ## DPU Mode
 
 When running with a DPU (Data Processing Unit) cluster, the `validate_offload` plugin needs
--- a/pluginPingMgmtPort.py
+++ b/pluginPingMgmtPort.py
@@ -1,0 +1,136 @@
+import ipaddress
+import json
+
+from typing import Any
+from typing import Optional
+
+from ktoolbox import common
+
+import pluginbase
+import task
+import tftbase
+
+from task import PluginTask
+from task import TaskOperation
+from testSettings import TestSettings
+from tftbase import BaseOutput
+from tftbase import PluginOutput
+from tftbase import TaskRole
+
+logger = common.ExtendedLogger("tft." + __name__)
+
+PING_COUNT = 5
+PING_TIMEOUT_SEC = 2
+
+
+def mgmt_port_ip_from_node_subnets(node_subnets_raw: str) -> str:
+    node_subnets = json.loads(node_subnets_raw)
+    subnet = node_subnets["default"][0]
+    network = ipaddress.ip_network(subnet, strict=False)
+    return str(network.network_address + 2)
+
+
+class PluginPingMgmtPort(pluginbase.Plugin):
+    PLUGIN_NAME = "ping_mgmt_port"
+
+    def _enable(
+        self,
+        *,
+        ts: TestSettings,
+        perf_server: task.ServerTask,
+        perf_client: task.ClientTask,
+        tenant: bool,
+    ) -> list[PluginTask]:
+        return [
+            TaskPingMgmtPort(ts, tenant),
+        ]
+
+
+plugin = pluginbase.register_plugin(PluginPingMgmtPort())
+
+
+class TaskPingMgmtPort(PluginTask):
+    @property
+    def plugin(self) -> pluginbase.Plugin:
+        return plugin
+
+    def __init__(self, ts: TestSettings, tenant: bool):
+        super().__init__(
+            ts=ts,
+            index=0,
+            task_role=TaskRole.CLIENT,
+            tenant=tenant,
+        )
+
+        self.pod_name = f"tools-pod-{self.node_location}-client-ping-mgmt-port"
+        self.in_file_template = tftbase.get_manifest("tools-pod.yaml.j2")
+
+    def initialize(self) -> None:
+        super().initialize()
+        self.render_pod_file("Plugin Pod Yaml")
+
+    def _get_mgmt_port_ip(self) -> str:
+        server_node = self.ts.node_server.name
+
+        node_data = self.run_oc_get(f"node/{server_node}", namespace=None)
+        if node_data is None:
+            raise RuntimeError(f"Management port node {server_node!r} not found")
+
+        annotations = node_data.get("metadata", {}).get("annotations", {})
+        node_subnets_raw = annotations.get("k8s.ovn.org/node-subnets")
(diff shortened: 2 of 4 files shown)


Venkat-Kunaparaju:
Adding ping as a new test type makes sense, but one issue I have is this new test case will only work with ping. Currently if we specify `iperf-tcp` or any other connection type, we are able to run all the test cases in the same suite because they all operate independent of connection type. This won't be the case anymore with this new test. Not sure how we can go about this. WDYT @wizhaoredhat ?

wizhaoredhat:
> Adding ping as a new test type makes sense, but one issue I have is this new test case will only work with ping. Currently if we specify `iperf-tcp` or any other connection type, we are able to run all the test cases in the same suite because they all operate independent of connection type. This won't be the case anymore with this new test. Not sure how we can go about this. WDYT @wizhaoredhat ?

Good catch @Venkat-Kunaparaju . This test is out of the norm since it is not talking between pods / containers. So it is quite awkward fitting in the current framework. However it is effeminately a useful test to test connectivity to the management ports. I feel like this might fit as a plugin that can be optionally turn on while we run the regular test cases. Like a healthcheck plugin. WDYT @Venkat-Kunaparaju / @josecastillolema . At the same level as measure_power, etc...

Venkat-Kunaparaju:
> > Adding ping as a new test type makes sense, but one issue I have is this new test case will only work with ping. Currently if we specify `iperf-tcp` or any other connection type, we are able to run all the test cases in the same suite because they all operate independent of connection type. This won't be the case anymore with this new test. Not sure how we can go about this. WDYT @wizhaoredhat ?
> 
> Good catch @Venkat-Kunaparaju . This test is out of the norm since it is not talking between pods / containers. So it is quite awkward fitting in the current framework. However it is effeminately a useful test to test connectivity to the management ports. I feel like this might fit as a plugin that can be optionally turn on while we run the regular test cases. Like a healthcheck plugin. WDYT @Venkat-Kunaparaju / @josecastillolema . At the same level as measure_power, etc...

I like this direction better. What I think we should do is something I thought about before, but didn't think there was a wide enough use case for. 

We should have a new field in config.yaml called something like `health_checks` that operates very similar to `plugins`, only that it doesn't run alongside a specific test case, but instead only runs once per `tft` entry. But the rest of the characteristics of `plugins` should still be there: it's functionality is entirely independent of the structure of the test cases where there are a lot of rules to conform to and all `health_checks` will be their own .py file, just like how plugins are so they can be as flexible and do whatever needs to be done independent of anything else. They can still read config.yaml values like server and client and pull that information as needed. We would need to update the print results section to include this new field. 

So how this relates to this PR is that this PR can be implemented as a `health_check`. It doesn't fit exactly into the `plugin` space IMO because it is still independent of a test case. Unlike measure_power or measure_cpu which need a test case to function. The new PING test type is not needed and ping can be hard coded into the logic of this health_check. So like @wizhaoredhat mentioned, this avoids the need to conform into the framework of test cases and kind of have free reign over what it can do.

There are other potential `health_checks` that can be added in the future that might prove useful which is why I think we should add this not only to support this PR. Stuff like pod->internet (ex. ping google.com), pod->dns. There are probably also environment specific endpoints that users want to test against and can easily implement (which is the case for us). Also flow-test adjacent stuff like maybe mtu checks of ports, vf checks on pod, etc. This way we can expand the scope of flow tests and support not only the current framework of flow tests. The test cases will still be the core, since ultimately it is the most important thing to check for,  but having health checks to run before the test cases can be beneficial. 

I know I said a lot so let me know what you guys think 😄. What I proposed might be out of scope of this PR currently.

wizhaoredhat:
@Venkat-Kunaparaju  I like this new health_check section even better than the plugin approach. I am not sure if @josecastillolema would have cycles to work on it (knowing his schedule).

Venkat-Kunaparaju:
@wizhaoredhat I think for now we can implement it as a plugin and run it that way, but in the future I do plan to add some pod->internet test and when that comes along I can implement this feature I just talked about and migrate pod->mgmt_port from a plugin to a health_check. If someone wants to implement health_check sooner than they are welcome to. As long as @josecastillolema is okay with changing how we call this test now (via plugin) and in the future (via health_check).

josecastillolema:
Reimplemented as a plugin, let me know your thoughts @Venkat-Kunaparaju @wizhaoredhat :
```yaml
tft:
  - name: "Mgmt Port Plugin Test"
    namespace: "tft-mgmt-port-plugin-test"
    test_cases: "2"
    duration: "5"
    connections:
      - name: "Connection_1"
        type: "iperf-tcp"
        instances: 1
        server:
          - name: "worker1.example.com"
        client:
          - name: "master1.example.com"
        plugins:
          - ping_mgmt_port
kubeconfig: "~/.kube/config"
kubeconfig_infra:
```

```sh
python ./tft.py tft-mgmt-port-plugin-test.yaml
...
2026-07-11 12:00:16.417 INFO    [th:140390339422080]: Ping to management port 10.131.0.2 on node worker1.example.com succeeded
...
2026-07-11 12:00:46.082 INFO    [th:140390339422080]:      plugin ping_mgmt_port (client), succeeded
...
```

Venkat-Kunaparaju:
I think we need to rebase this to get the CI lanes to pass. It looks like it is using an old tft config that is resulting in the failure. Rebasing should update it.

josecastillolema:
> I think we need to rebase this to get the CI lanes to pass. It looks like it is using an old tft config that is resulting in the failure. Rebasing should update it.

@Venkat-Kunaparaju fixed now I think, all CI is green

pluginPingMgmtPort.py
```diff
@@ -0,0 +1,136 @@
...
+    def _get_mgmt_port_ip(self) -> str:
+        server_node = self.ts.node_server.name
+
+        node_data = self.run_oc_get(f"node/{server_node}", namespace=None)
+        if node_data is None:
+            raise RuntimeError(f"Management port node {server_node!r} not found")
+
+        annotations = node_data.get("metadata", {}).get("annotations", {})
+        node_subnets_raw = annotations.get("k8s.ovn.org/node-subnets")
+        if node_subnets_raw is None:
```
Venkat-Kunaparaju, line 81: I think if the management port IP is missing or can't be obtained, we shouldn't hard terminate the whole tft test suite. We should only scope it to fail the plugin. This will lower the blast radius of one specific plugin/test case affecting the output of the others.

Venkat-Kunaparaju (changes requested):

pluginPingMgmtPort.py
josecastillolema: Fixed by catching `RuntimeError` from `_get_mgmt_port_ip()` inside the thread action and returning a failed `PluginOutput`, so a missing/invalid management port IP only fails the plugin instead of terminating the entire test suite.

README.md
Venkat-Kunaparaju, line 181: I think its okay for now if we don't include this plugin here, since it doesn't fit with the rest of the plugins. The goal is to replace the structure of this in the future anyways.

Venkat-Kunaparaju:
lgtm, thanks for the updates

README.md
josecastillolema: But once we merge this and until we replace the structure in the future it will be a plugin right?
Could we delete this once we get the restructure in place in the future and keep it for now?

Venkat-Kunaparaju: Yeah sure, that works with me as well.

josecastillolema: Thanks!

Venkat-Kunaparaju (approved):

PR #311, opened 2026-07-08
Use mypy version 2.1.0 instead of latest version

Venkat-Kunaparaju:
Mypy 2.2.0 leading to failures for very old code

## Summary by CodeRabbit

* **Chores**
  * Updated development dependency settings by removing an inherited requirements reference.
  * Pinned the type-checking tool to a specific version for more consistent development builds.

### DIFF
--- a/requirements-devel.txt
+++ b/requirements-devel.txt
@@ -1,7 +1,7 @@
 -r requirements.txt
 black
 flake8
-mypy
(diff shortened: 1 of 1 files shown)


wizhaoredhat (approved):
LGTM. Will investigate later how to use the latest mypy

PR #314, opened 2026-07-14
Add pod startup diagnostics and reduce readiness timeout

Venkat-Kunaparaju:
Capture apply and readiness failures, dump the pod description, and raise an explicit setup error.

Reduce the readiness wait from 10 minutes to 2 minutes so pod startup failures are reported sooner.

10 minutes is a long time to wait for pod bringup, and if there is a failure in pod bringup, the user has to wait the full time for failure. Reducing to 2 minutes, which is a more reasonable timeframe.

If the pod fails to be brought up, the test fails and cleans up all resources by default. This prevents the user from inspecting why the pod bringup failed. Printing out pod description will help in debugging why the pod didn't get brought up without rerunning the tests and inspecting the logs manually. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved diagnostics when application pod creation fails.
  * Added clearer error details when a pod does not become ready within the two-minute startup window.
  * Pod startup failures now return actionable error messages instead of waiting up to 10 minutes.

### DIFF
--- a/task.py
+++ b/task.py
@@ -1020,15 +1020,35 @@ class Task(ABC):
         v = self.run_oc_get(f"pod/{self.pod_name}", may_fail=True)
         if v is None:
             logger.info(f"Creating Pod {self.pod_name}.")
-            self.run_oc(f"apply -f {self.out_file_yaml}", die_on_error=True)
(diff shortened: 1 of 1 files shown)


wizhaoredhat:
@coderabbitai review

wizhaoredhat (approved):

PR #315, opened 2026-07-17
iperf: auto-detect IP family for ClusterIP, add ip_family config option

tsorya:
## Summary
- Auto-detect the IP family from the actual ClusterIP service address and pass `-4` or `-6` to iperf3 accordingly. This prevents "Bad file descriptor" errors caused by iperf3 attempting IPv6 on IPv4-only services (or vice versa).
- Add an `ip_family` config option (default: `"auto"`) allowing users to explicitly force `"ipv4"`, `"ipv6"`, or `"any"` (no flag) when needed.
- When `ip_family` is `"auto"` and connection mode is not `CLUSTER_IP`, no flag is passed (original behavior).

## Behavior
| `ip_family` | ClusterIP (IPv4 svc) | ClusterIP (IPv6 svc) | Other modes |
|---|---|---|---|
| `"auto"` (default) | `-4` | `-6` | no flag |
| `"ipv4"` | `-4` | `-4` | `-4` |
| `"ipv6"` | `-6` | `-6` | `-6` |
| `"any"` | no flag | no flag | no flag |

## Test plan
- [ ] Run HOST_TO_CLUSTER_IP_TO_POD_SAME_NODE iperf test on a cluster with IPv6 addresses on host interfaces and IPv4 ClusterIP
- [ ] Verify non-ClusterIP iperf tests still work without any flag
- [ ] Verify explicit `ip_family: "ipv6"` overrides auto-detection

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved iperf3 connectivity for ClusterIP connections by automatically selecting the correct iperf3 IPv4/IPv6 flag on both server and client.
  * Kept existing behavior for non-ClusterIP connection modes and for explicitly configured IP family values.
* **New Features**
  * Added optional `ip_family` configuration for `tft`, supporting `auto` (default), `ipv4`, `ipv6`, and `any`, with inline documentation.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -5,6 +5,11 @@ tft:
     # i.e. "1-9,15-19" or using the pretty names.
     test_cases: "1"
     duration: "30"
+    # ip_family: "auto" (default), "ipv4", "ipv6", or "any".
+    # Controls the -4/-6 flag passed to iperf3.
+    # "auto" detects the IP family from the ClusterIP service address;
+    # for non-ClusterIP connections it passes no flag.
+    # ip_family: "auto"
     # "logs": location of test results. Defaults to "<working-dir>/ft-logs/".
     # It is better to pass "--output-base" to ./tft.py.
     # "privileged_pod": whether to run test pods as privileged. Defaults to false.
--- a/testConfig.py
+++ b/testConfig.py
@@ -29,6 +29,7 @@ from ktoolbox.k8sClient import K8sClient
 from pluginbase import Plugin
 from testType import TestTypeHandler
 from tftbase import ClusterMode
+from tftbase import IpFamily
 from tftbase import PodType
 from tftbase import TestCaseType
 from tftbase import TestType
@@ -604,6 +605,7 @@ class ConfTest(StructParseBaseNamed):
     capabilities_pod: Mapping[str, tuple[str, ...]]
     connections: tuple[ConfConnection, ...]
     logs: pathlib.Path
+    ip_family: IpFamily
 
     def __post_init__(self) -> None:
         for c in self.connections:
@@ -624,6 +626,7 @@ class ConfTest(StructParseBaseNamed):
             "capabilities_pod": self.capabilities_pod,
             "connections": [c.serialize() for c in self.connections],
             "logs": str(self.logs),
+            "ip_family": self.ip_family.value,
         }
 
     @staticmethod
@@ -689,6 +692,12 @@ class ConfTest(StructParseBaseNamed):
                 default="ft-logs",
             )
 
+            ip_family = common.structparse_pop_enum(
+                varg.for_key("ip_family"),
+                enum_type=IpFamily,
+                default=IpFamily.AUTO,
+            )
+
         return ConfTest(
             yamlidx=pctx.yamlidx,
             yamlpath=pctx.yamlpath,
@@ -701,6 +710,7 @@ class ConfTest(StructParseBaseNamed):
             capabilities_pod=capabilities_pod,
             connections=connections,
             logs=pathlib.Path(logs),
+            ip_family=ip_family,
         )
 
     @property
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ tft:
     test_cases: "(3)"
     duration: "(4)"
     pre_provision: (5)
+    ip_family: "(33)"
     # Location of artifacts from run can be specified: default <working-dir>/ft-logs/
     # logs: "/tmp/ft-logs"
     connections:
@@ -203,7 +204,12 @@ kubeconfig_infra: (32)
 32. "kubeconfig", "kubeconfig_infra": if set to non-empty strings, then these are the KUBECONFIG
   files. "kubeconfig_infra" must be set for DPU cluster mode. If both are empty, the configs
   are detected based on the files we find at /root/kubeconfig.*.
-33. "dpu_node_host_label": (Required for DPU mode) The label on DPU nodes that identifies
+33. "ip_family" - (Optional) Controls the `-4`/`-6` flag passed to iperf3. Accepted values:
+     `"auto"` (default), `"ipv4"`, `"ipv6"`, `"any"`. When set to `"auto"`, the IP family is
+     detected from the actual ClusterIP service address for ClusterIP connection modes; for
+     other connection modes no flag is passed. Set explicitly to `"ipv4"` or `"ipv6"` to force
+     the IP family regardless of connection mode, or `"any"` to let iperf3 auto-detect.
+34. "dpu_node_host_label": (Required for DPU mode) The label on DPU nodes that identifies
   which host worker node they belong to. For NVIDIA DPUs, use `provisioning.dpu.nvidia.com/host`.
 
 
--- a/testTypeIperf.py
+++ b/testTypeIperf.py
@@ -1,3 +1,4 @@
+import ipaddress
 import json
 import shlex
 import task
@@ -18,7 +19,9 @@ from testSettings import TestSettings
 from testType import TestTypeHandler
 from tftbase import BaseOutput
 from tftbase import Bitrate
+from tftbase import ConnectionMode
 from tftbase import FlowTestOutput
+from tftbase import IpFamily
 from tftbase import TestType
 
 logger = common.ExtendedLogger("tft." + __name__)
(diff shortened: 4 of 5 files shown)


config.yaml
```diff
@@ -5,6 +5,10 @@ tft:
     # i.e. "1-9,15-19" or using the pretty names.
     test_cases: "1"
     duration: "30"
+    # "ip_family": "ipv4" (default), "ipv6", or "any".
+    # Forces iperf3 to use -4/-6. Use "ipv4" on clusters where hosts have
+    # IPv6 addresses but services are IPv4-only (e.g. DPU nodes).
+    ip_family: "ipv4"
```
Venkat-Kunaparaju: Since this is a config file change, please update the README as well. 

I think we should rename this field since its specific to iperf, so its not confusing for people running another test type (for example "http") and expecting ipv6. Maybe "ip_family_perf"?

testConfig.py
```diff
@@ -689,6 +692,12 @@ def parse(pctx: StructParseParseContext) -> "ConfTest":
                 default="ft-logs",
             )
 
+            ip_family = common.structparse_pop_enum(
+                varg.for_key("ip_family"),
+                enum_type=IpFamily,
+                default=IpFamily.IPV4,
```
Venkat-Kunaparaju: Shouldn't the default be `ANY` here instead of `IPV4`? Because it was the old functionality to auto detect the ip family for iperf to use.

wizhaoredhat:
@Venkat-Kunaparaju Is this related to the SERVICE_NAME change you did recently?

Venkat-Kunaparaju:
> @Venkat-Kunaparaju Is this related to the SERVICE_NAME change you did recently?

Yes, I guess using DNS name for ClusterIP in this case decides to use IPv6 instead of IPv4 for iperf, because iperf doesnt know which one to exactly use since its reference is a DNS name, not an ip family like before.

testConfig.py
tsorya: Default was ipv4 till now as far as I understand

wizhaoredhat: I think we can fix this a different way. Let's hardcode when using ClusterIP to uses "-4" option for iperf because ClusterIP only supports IPv4 (atm). Then we don't need to configure it globally.
WDYT?

Venkat-Kunaparaju: @tsorya Before iperf would automatically detect whether to use IPv4 or IPv6 based on the IP address or hostname you provide. If you specify a hostname that resolves to an IPv6 address, it uses IPv6; if you provide an IPv4 address, it uses IPv4.

@wizhaoredhat If clusterip only supports IPv4, then that fix works with me.

tsorya:
@wizhaoredhat @Venkat-Kunaparaju fixed as you asked

testConfig.py
wizhaoredhat: I haven't played too much with IPv6 yet. So I myself is unsure.

Venkat-Kunaparaju:
LGTM as long as we can confirm that clusterip currently only support ipv4.

tsorya:
> LGTM as long as we can confirm that clusterip currently only support ipv4.

@wizhaoredhat @Venkat-Kunaparaju  as far as i checked cluster ip supports ipv6.

I updated the PR to automatically handle it, if cluster is ipv4 use ipv4 and added an option to override it

Venkat-Kunaparaju:
I just noticed that we already support passing custom iperf options through server.args and client.args. For this case, we could set -4 in the config file to solve the ClusterIP via SERVICE_NAME issue:
```
server:
  - args: "-4"
client:
  - args: "-4"
```
So in this way existing iperf behavior would remain unchanged. Would this cover the original issue you were facing without introducing the additional `ip_family` configuration feature?

tsorya:
> I just noticed that we already support passing custom iperf options through server.args and client.args. For this case, we could set -4 in the config file to solve the ClusterIP via SERVICE_NAME issue:
> 
> ```
> server:
>   - args: "-4"
> client:
>   - args: "-4"
> ```
> 
> So in this way existing iperf behavior would remain unchanged. Would this cover the original issue you were facing without introducing the additional `ip_family` configuration feature?

We can remove the option but someone broke default behavior and broke CIs, there is no point to use ipv6 when cluster is running with ipv4.
So IMO automating the ip family choice sounds right to me. 
Another way we will need to update already running CIs

Venkat-Kunaparaju:
> > I just noticed that we already support passing custom iperf options through server.args and client.args. For this case, we could set -4 in the config file to solve the ClusterIP via SERVICE_NAME issue:
> > ```
> > server:
> >   - args: "-4"
> > client:
> >   - args: "-4"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So in this way existing iperf behavior would remain unchanged. Would this cover the original issue you were facing without introducing the additional `ip_family` configuration feature?
> 
> We can remove the option but someone broke default behavior and broke CIs, there is no point to use ipv6 when cluster is running with ipv4. So IMO automating the ip family choice sounds right to me. Another way we will need to update already running CIs

So the thing is, iperf is already designed to automate the ip family choice. If we don't specify "-4" or "-6", iperf will look at the ip family of the destination/server ip, and determine what family to use for the iperf connection. 

The change I made which affected this was that we are using `SERVICE_NAME` (service DNS name) for clusterIP instead of its actually ipv4 address in this PR: https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/297. Me and @wizhaoredhat decided to use `SERVICE_NAME` for the default for service based connections like clusterIP was that it would also help catching dns issues in addition to testing connectivity. 

I agree, there is no point using ipv6 instead of ipv4 if the cluster is indeed configured for ipv4. The change I made does not affect this. If the cluster is configured only for ipv4, then dns should pick up the ipv4 address of the server. I tested this out in my own cluster configured for ipv4 and dns correctly resolves the clusterip to an ipv4 address, so the iperf connection uses ipv4. This means that is a cluster specific issue and not something necessarily tied to TFT. 
https://github.com/ovn-kubernetes/ovn-kubernetes/actions/runs/29767774263/job/88438054233?pr=6707
here is an example of a kind cluster configured for v4 with clusterIP `SERVICE_NAME` passing with iperf3 on ipv4.

What I think is happening is that DNS in the CI you are referring to (not sure if I have access) is, I believe, resolving to an ipv6 address for the clusterip dns name, which then results in the iperf connection using ipv6. This is because iperf is designed to switch between ipv6 or ipv4 depending on the destination ip, which DNS resolved to an ipv6 address. But from what I can see is that with this change we can force DNS to resolve to a specific ip family with "-4", "-6".

I am not sure exactly why DNS is resolving to an ipv6 address, but if DNS components in an ipv4 cluster have the ability to resolve either ipv4 and ipv6 address for services configured for ipv4, then I think this feature is needed. I am not sure if this is the case though I haven't really played with k8s DNS 😅. Either way - to restore the default behavior you can always set `TFT_DEFAULT_TARGET_ACCESS_MODE=IP` and run the tests so that you can avoid using `SERVICE_NAME` and use the direct IP for the service. Or you can use the "args" approach I mentioned above as well. By default, TFT_DEFAULT_TARGET_ACCESS_MODE is set to `SERVICE_NAME`, but we can change it back to `IP` as well.

If my understanding is wrong anywhere, please let me know.

@wizhaoredhat let me know what you think as well.

tsorya:
The issue is that your change changed default behavior, it means that we need to change our jobs in order to restore it.
In general I would prefer to drop this change but restore previous default to use ip

Venkat-Kunaparaju:
Yeah, we can restore the previous behavior to use ip instead of service name for services. Will put in a PR soon and lmk if  the fix will work.

Venkat-Kunaparaju:
> Yeah, we can restore the previous behavior to use ip instead of service name for services. Will put in a PR soon and lmk if the fix will work.

https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/316 Change here

wizhaoredhat:
@tsorya Unfortunately I believe your fix here doesn't work all the time. I tested it on the doca cluster and it fails. I tried to debug why but I believe it has something to do with Hardware Offloading with the DPU. Something with the DPU gateway is malfunctioning and I can't reproduce with DPU Simulation (No hardware offloading). E.g. Here by adding a global IPv6 addresss: https://github.com/ovn-kubernetes/dpu-simulator/pull/34

tsorya:
I closed it, fix provided by @Venkat-Kunaparaju works for us

PR #316, opened 2026-07-20
Default service access point to IP instead of SERVICE_NAME

Venkat-Kunaparaju:
## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default target access mode for service-related connections to use service IP addresses.
  * Preserved support for explicitly selecting service DNS names through the environment variable.

* **Documentation**
  * Corrected the documented default for `TFT_DEFAULT_TARGET_ACCESS_MODE` to `IP`.
  * Clarified the available values and their behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -444,8 +444,8 @@ match. The `EgressIP` resource and the egress node's labels are removed during c
 - `TFT_KUBECONFIG`, `TFT_KUBECONFIG_INFRA` to overwrite the kubeconfigs from the configuration
      file. See also the "--kubeconfig" and "--kubeconfig-infra" command line options.
 - `TFT_DEFAULT_TARGET_ACCESS_MODE` controls the normal target access mode for service-backed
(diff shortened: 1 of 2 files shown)


Venkat-Kunaparaju:
@tsorya @wizhaoredhat PTAL

tsorya:
--- Failing Flows (0 in '/root/doca8/ci/openshift-dpf-2026-07-20_23-41-01/openshift-dpf/scripts/../repos/kubernetes-traffic-flow-tests/results/20260721_230018000.json') ---


No failures detected in results

nice !!!

Venkat-Kunaparaju:
After some investigation, it looks host->DNS is a bug with OVS-DOCA. Once there is a new release for OVS-DOCA that has a fix for it, the default access method for services will fall back to IP. 

At some point in the future, we should switch it back to SERVICE_NAME once we know it will be stable across different environments.

wizhaoredhat:
Sorry, forgot to LGTM and merge. RH is using DOCA 3.4.0 which has a bug.
Let's set the default behavior to use IP.

PR #317, opened 2026-07-21
Make pod bring-up timeout configurable

Venkat-Kunaparaju:
Add TFT_POD_BRINGUP_TIMEOUT to control how long TFT waits for pods to become ready. Keep 2m as the default and use the configured value in readiness logs and errors.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable timeout for waiting for pods to become ready.
  * Supports Kubernetes duration formats such as `30s` and `5m`.
  * Defaults to a two-minute timeout when no value is configured.

* **Bug Fixes**
  * Timeout-related readiness errors now report the configured duration for clearer diagnostics.

* **Documentation**
  * Documented the new `TFT_POD_BRINGUP_TIMEOUT` environment variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -466,6 +466,8 @@ match. The `EgressIP` resource and the egress node's labels are removed during c
 - `TFT_HOST_NETWORK_NAMESPACE` the namespace used as the `namespaceSelector` target for the
      `HOST_TO_POD_NP_NS_SELECTOR_ALLOW` test case. Defaults to `ovn-host-network`.
      For OpenShift clusters, it must be set to `openshift-host-network`.
+- `TFT_POD_BRINGUP_TIMEOUT` controls how long TFT waits for a pod to become ready. Accepts a
+     Kubernetes duration such as `30s` or `5m`. Defaults to `2m`.
 
 ## File Transfer via magic-wormhole
 
--- a/task.py
+++ b/task.py
@@ -1029,14 +1029,17 @@ class Task(ABC):
         else:
             logger.info(f"Pod {self.pod_name} already exists.")
 
-        logger.info(f"Waiting for Pod {self.pod_name} to become ready.")
+        timeout = tftbase.get_tft_pod_bringup_timeout()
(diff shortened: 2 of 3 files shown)


Venkat-Kunaparaju:
This is a follow up to: https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/314

This will preserve existing behavior if it is required for some environments when pod bring up is slow, but also not affect faster environments.

wizhaoredhat:
Why make this configurable? If we extend the time, it's really a performance bug that shouldn't be overlooked.
Shortening the time does nothing because if it succeeds it doesn't wait the full 2 minutes.

Venkat-Kunaparaju:
> Why make this configurable? If we extend the time, it's really a performance bug that shouldn't be overlooked. Shortening the time does nothing because if it succeeds it doesn't wait the full 2 minutes.

That is a good point. It's more so like where the environment is already known and accepted to have some performance bottleneck. For example, a simulated environment with some slow network will take a long time for the test pod image pull. Or kubelet itself takes a while setting up the pod, etc. The time for each of these isnt the same for each environment, so allowing the user to adjust the time for pod bring up is useful to adjust the headroom.

But the reason to shorten the time in the first place is mainly to catch unpredicted issues sooner. For example if you expect image pull to happen quick, then the time can be shortened, and if the time gets exceeded, then the user knows its a true error for them. Additionally, sometimes the known issues can be something like not enough vfs available to setup the pod, so it will sit in an idle state, never becoming running, then you have to wait the full default timeout to inspect the issue when it could have been revealed sooner. 

Its similar to how we allow support for adjusting the iperf throughput threshold, there isn't a single number that will be universal throughout different environments IMO.

wizhaoredhat:
Hmm ok, fair points. I would want to use kube-burner first to get the pod latency. I wouldn't want the tft tools to do the work of kube-burner. But I can see lowering it to 5 seconds (if we expect the worst case pod latency to be 5 seconds and anything more would be bad), for example to set thresholds.

wizhaoredhat (approved):

PR #320, opened 2026-08-06
Add TFT_EXTERNAL_SERVER to skip local podman

josecastillolema:
Adds `TFT_EXTERNAL_SERVER` to skip local podman and use a pre-existing external server.

- [x] Feature
- [x] Documentation update
- [x] Unit test
- [x] Manually tested

Need this to be able to run from Prow, cc @wizhaoredhat 

Assisted-by: Claude 🤖

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring an external test server through the `TFT_EXTERNAL_SERVER` environment variable.
  * Supports hostnames, IPv4 addresses, optional ports, and IPv6 address formats.
  * External tests can now connect to the configured server and port, with sensible port fallback behavior.

* **Documentation**
  * Documented supported address formats, defaults, applicable test modes, and a Podman setup example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -491,7 +491,7 @@ match. The `EgressIP` resource and the egress node's labels are removed during c
 - `TFT_EXTERNAL_SERVER` address of a pre-existing external server in `host[:port]` format
      (e.g. `192.168.1.100:5201`). Works with any test type (iperf, netperf, http).
      When set and the connection mode is `EXTERNAL_IP` (`POD_TO_EXTERNAL`, `HOST_TO_EXTERNAL`,
-     `UDN_PRIMARY_POD_TO_EXTERNAL`, `POD_TO_EXTERNAL_EGRESS`), no local Podman container is
+     `UDN_PRIMARY_POD_TO_EXTERNAL`), no local Podman container is
      started; the client pod connects directly to the specified server. If port is omitted,
      the configured `pod_port` is used (defaults to `5201`). IPv6 addresses use bracket notation
      (e.g. `[fd00::1]:5201`). To start an iperf3 server on the remote host:
--- a/README.md
+++ b/README.md
@@ -488,6 +488,16 @@ match. The `EgressIP` resource and the egress node's labels are removed during c
 - `TFT_CUDN_LOCALNET_PHYSICAL_NETWORK` physical network name for localnet CUDN tests. Defaults to `physnet`.
 - `TFT_UDN_NO_OVERLAY_OUTBOUND_SNAT_ENABLED` outbound SNAT setting for no-overlay CUDNs. Defaults to `true`.
 - `TFT_UDN_NO_OVERLAY_ROUTING_MANAGED` whether OVN-Kubernetes manages routing for no-overlay CUDNs. Defaults to `false` (unmanaged).
+- `TFT_EXTERNAL_SERVER` address of a pre-existing external server in `host[:port]` format
+     (e.g. `192.168.1.100:5201`). Works with any test type (iperf, netperf, http).
+     When set and the connection mode is `EXTERNAL_IP` (`POD_TO_EXTERNAL`, `HOST_TO_EXTERNAL`,
+     `UDN_PRIMARY_POD_TO_EXTERNAL`, `POD_TO_EXTERNAL_EGRESS`), no local Podman container is
+     started; the client pod connects directly to the specified server. If port is omitted,
+     the configured `pod_port` is used (defaults to `5201`). IPv6 addresses use bracket notation
+     (e.g. `[fd00::1]:5201`). To start an iperf3 server on the remote host:
+     ```
+     podman run --rm -p 5201:5201 ghcr.io/ovn-kubernetes/kubernetes-traffic-flow-tests:latest iperf3 -s -p 5201
+     ```
 - `TFT_EXTERNAL_URL` URL to curl for external connectivity tests (e.g. `http://google.com`).
      Only effective when the connection type is `http` and the connection mode is `POD_TO_EXTERNAL`
      or `HOST_TO_EXTERNAL`. When set, no Podman server is started; the client pod curls this URL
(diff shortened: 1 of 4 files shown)


task.py
```diff
@@ -1468,8 +1469,7 @@ def _create_setup_operation(
 
         self.pre_provisioning = provisioning
 
-        if self.use_internet:
-            # No server needed; client curls external URL directly.
+        if self.use_external:
             self.ts.event_server_alive.set()
             return None
```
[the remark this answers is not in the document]
josecastillolema: Correct me if I am wrong but I think the code already handles it correctly. At task.py:1842, when server_stdout is None, the test is explicitly marked as failed with the message "EgressIP verification failed: no server output captured". It does NOT silently return success.

So this is a non-issue — no fix needed. The existing code already catches the case where use_external is True and server_stdout is unset: the egress verification fails explicitly rather than silently passing.

task.py
```diff
@@ -1774,9 +1776,13 @@ def get_target_ip(self) -> str:
...
-        """Get the port to connect to. For EXTERNAL_IP, use discovered external_port."""
-        if self.server.use_internet:
-            # URL is used directly by testTypeHttp; no port needed.
+        if self.server.use_external:
+            external_server = tftbase.get_tft_external_server()
+            if external_server is not None:
+                _, port = external_server
+                if port is not None:
+                    return port
+                return self.server.port
```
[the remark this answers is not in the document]
josecastillolema: Should be fixed in https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/320/changes/9c35549ab18adf987050a1849c3247c357008739

README.md
```diff
@@ -488,6 +488,16 @@ match. The `EgressIP` resource and the egress node's labels are removed during c
...
+- `TFT_EXTERNAL_SERVER` address of a pre-existing external server in `host[:port]` format
+     (e.g. `192.168.1.100:5201`). Works with any test type (iperf, netperf, http).
+     When set and the connection mode is `EXTERNAL_IP` (`POD_TO_EXTERNAL`, `HOST_TO_EXTERNAL`,
+     `UDN_PRIMARY_POD_TO_EXTERNAL`, `POD_TO_EXTERNAL_EGRESS`), no local Podman container is
+     started; the client pod connects directly to the specified server. If port is omitted,
+     the configured `pod_port` is used (defaults to `5201`). IPv6 addresses use bracket notation
+     (e.g. `[fd00::1]:5201`). To start an iperf3 server on the remote host:
+     ```
+     podman run --rm -p 5201:5201 ghcr.io/ovn-kubernetes/kubernetes-traffic-flow-tests:latest iperf3 -s -p 5201
+     ```
```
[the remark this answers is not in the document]
josecastillolema: Should be fixed in https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/320/changes/9c35549ab18adf987050a1849c3247c357008739

task.py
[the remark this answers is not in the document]
josecastillolema: Should be fixed in https://github.com/ovn-kubernetes/kubernetes-traffic-flow-tests/pull/320/commits/d5cf2133cdb427dc260bea08b8a0fa0ad4cec981

task.py
```diff
@@ -1261,13 +1261,22 @@ def __init__(self, ts: TestSettings):
...
+        use_external = connection_mode == ConnectionMode.EXTERNAL_IP and (
+            tftbase.get_tft_external_server() is not None
+            or (
+                tftbase.get_tft_external_url() is not None
+                and ts.connection.test_type == TestType.HTTP
+            )
         )
 
-        if use_internet:
+        if use_external and ts.connection.has_egress_ip:
```
Venkat-Kunaparaju, line 1272: This means that if a connection specifies the egress IP test case when there are other test cases as well, then the whole connection is rendered invalid. 

I think instead of this, we should scope `use_external` to something like 

```
use_external = (
    connection_mode == ConnectionMode.EXTERNAL_IP
    and not ts.test_case_id.is_egress_ip
    and (
        tftbase.get_tft_external_server() is not None
        or (
            tftbase.get_tft_external_url() is not None
            and ts.connection.test_type == TestType.HTTP
        )
    )
)
```
This basically means we will use the server configured via the env variable for other EXTERNAL_IP tests except for egress IP since it won't work with it since we can't inspect the server.

WDYT?

wizhaoredhat: I think this is a good suggestion, this means we don't error out with "TFT_EXTERNAL_SERVER cannot be used with EgressIP tests because the external server's output is not available for source-IP verification", right?

Venkat-Kunaparaju: Yep exactly

josecastillolema: But that would mean that egress IP tests fall back to the local podman server?
I personally find it confusing, the user is setting `EXTERNAL_IP` and it's getting a test running on local podman container? IMO failing is much consistent UX.
Plus, one of the main consumers of this var will be [Prow](https://docs.prow.k8s.io/docs/overview/) where the test runs from a pod and podman is not available ...
Let me know your thoughts

wizhaoredhat: Hmmm no that is not my understanding. I wouldn't want egress IP tests to fall back to the local podman server at all when use_external (before use_internet) is true. 

Why would there be a test running on local podman container when EXTERNAL_IP is set?

Venkat-Kunaparaju: In the example I gave above, the local podman server is used if the test case id is egress IP even if EXTERNAL_IP is set. This is because EXTERNAL_IP won't be supported with egress IP, so instead of failing, we can still run the egress IP test. But other tests that can use EXTERNAL_IP will use it, its just for egress IP we will not use it and use the local podman server, because its test case scoped.

So if we specify POD_TO_EXTERNAL and POD_TO_EXTERNAL_EGRESS, and we set EXTERNAL_IP, then POD_TO_EXTERNAL will use EXTERNAL_IP and POD_TO_EXTERNAL_EGRESS will use the local podman server, instead of hard failing since POD_TO_EXTERNAL_EGRESS and EXTERNAL_IP are set when POD_TO_EXTERNAL and EXTERNAL_IP is still a valid case.

I dont feel too strongly about this, so if anyone disagrees we can stick to the current implementation.

wizhaoredhat: I think the current implementation is ok. I wouldn't want to mix an external server + local podman server together. IMO, it's either or, not both.

wizhaoredhat (approved):
LGTM

PR #322, opened 2026-08-11
Log test commands with task results

Venkat-Kunaparaju:
We already run the server command, so we should also mimic the same with the client. This will also help with quick debugging perf failures because currently we only emit the command ran for the client if there is a connection failure not a throughput failure.

## Summary by CodeRabbit

* **Enhancements**
  * Task output now includes the executed command alongside the result status for flow test results.


wizhaoredhat:
LGTM

wizhaoredhat (approved):

PR #325, opened 2026-08-12
Preserve UDN namespace during resource cleanup

Venkat-Kunaparaju:
With the regular namespace, it is not deleted at the beginning of the run, it just cleans up resources in them. We should mimic the same behavior with {ns}-udn as well, because currently it tried to delete the udn ns at the beginning of the run if it already exists.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stale resource cleanup by removing TFT-labeled resources while preserving the existing namespace.
  * Updated cleanup logging to provide clearer, more accurate messages.


wizhaoredhat (approved):
LGTM

